PR #4, opened 2021-06-30
First changes made for testing at NREL

jrood-nrel:
I'm looking through and writing a list of what I think we will need to iron out. I'm sure there's more than this:

1. Using `~/.spack/downloads` for archives
2. Base `packages.yaml` preferences
3. I prefer minimal use of homebrew to avoid customization
4. I don't know why I have a custom hypre

Also, I copied over some stuff I currently use in the `build-test` repo while I'm converting from there to here.


configs/base/packages.yaml
```diff
@@ -1,26 +1,30 @@
...
     version: [1.12.2]
   tioga:
     version: [develop]
+  yaml-cpp:
+    version: [0.6.3]
   trilinos:
     version: [develop]
-    variants: ~adios2~alloptpkgs~amesos+amesos2~anasazi~aztec+belos+boost~cgns~chaco~complex~debug~dtk~epetra~epetraext+exodus+explicit_template_instantiation~float~fortran~fortrilinos+glm+gtest+hdf5~hypre~ifpack+ifpack2~intrepid~intrepid2~isorropia+kokkos~matio~mesquite+metis~minitensor~ml+mpi+muelu~mumps+netcdf~nox~openmp~phalanx~piro+pnetcdf~python~rol~rythmos~sacado+shards~shylu+stk~stratimikos~suite-sparse~superlu~superlu-dist~teko~tempus+teuchos+tpetra~x11~xsdkflags+zlib+zoltan+zoltan2 cxxstd=14 build_type=Release
+    variants: ~adios2~alloptpkgs~amesos+amesos2~anasazi~aztec+belos+boost~cgns~chaco~complex~debug~dtk~epetra~epetraext+exodus+explicit_template_instantiation~float~fortran~fortrilinos+glm+gtest+hdf5~hypre~ifpack+ifpack2~intrepid~intrepid2~isorropia+kokkos~matio~mesquite+metis~minitensor~ml+mpi+muelu~mumps+netcdf~nox~openmp~phalanx~piro+pnetcdf~python~rol~rythmos~sacado+shards~shylu+stk~stratimikos~suite-sparse+superlu~superlu-dist~teko~tempus+teuchos+tpetra~x11~xsdkflags+zlib+zoltan+zoltan2 cxxstd=14 build_type=Release
```
psakievich: I thought you got rid of the superlu dependency?

env-templates/darwin.yaml
```diff
@@ -1,61 +1,22 @@
...
+  - amr-wind-nightly +netcdf+hypre+masa+openfast~latest_amrex
+  - nalu-wind-nightly +fftw+hypre+openfast+tioga+wind-utils 
+  #- exawind-nightly
   packages:
     all:
-      target: [x86_64]
-      compiler: [apple-clang, gcc, clang]
-      variants: build_type=Release +mpi
+      #target: [x86_64]
+      compiler: [apple-clang@12.0.0, gcc, clang]
```
psakievich: I'd rather leave the compiler version and `host_name` out of a `darwin.yaml` template.  I'd like these templates to be relatively insensitive to who ever uses them.  So if someone with mojave wants to utilize the they will have to edit the compiler.  I'd rather have it be the other way around (i.e. if you have multiple apple-clang's then you can edit to restrict it to the one you want)

repos/exawind/packages/amr-wind-developer/package.py
```diff
@@ -1,20 +1,35 @@
 from spack import *
-from spack.pkg.builtin.amr_wind import AmrWind as bAmrWind
+from spack.pkg.exawind.amr_wind import AmrWind as bAmrWind
```
psakievich: Out of curiosity, what's currently missing in the builtin amr-wind package?

repos/exawind/packages/exawind-developer/package.py
```diff
@@ -0,0 +1,37 @@
+from spack import *
+from spack.pkg.exawind.exawind import Exawind as bExawind
+import os
+
+
+class ExawindDeveloper(bExawind):
+    variant('asan', default=False,
+            description='Turn on address sanitizer')
+    variant('compile_commands', default=False,
```
psakievich: Not necessary for this PR, but I've been thinking we should make `compile_commands` default to True for the developer packages.  I always want them on and forget them half the time.

psakievich (commented):
I like most of these things. Thanks for putting so many packages together and cleaning so many details up. I'm glad we're iterating on this together.

configs/base/packages.yaml
jrood-nrel: Oops yes I fixed that.

env-templates/darwin.yaml
jrood-nrel: Yeah, I wasn't really sure where to put the hostname, so I had it there for now.

repos/exawind/packages/amr-wind-developer/package.py
jrood-nrel: It uses a `AMR_WIND_ENABLE_FORTRAN` option that doesn't exist and adds a warning to cdash.

jrood-nrel:
Thanks for looking at this. Let's resolve anything we shouldn't merge here. I'm going to work on this more today and try to clean some things up now that I have some testing working mostly how I want.

psakievich:
@jrood-nrel here's my opinion on using `~/.spack/downloads` for archives.  I try to avoid it because I sometimes run out of home space, and I often delete `~/.spack`.  I also like to have it in the `spack-manager` directory so if multiple admins are going to build stuff we don't have to keep re-downloading the same archives. But that said, I'm fine with it for darwin, and we can always change the defaults for the machine specific configs. So if you want to make it the default I can just override it in the SNL specific platforms.

jrood-nrel:
Ok maybe this is a little bit more digestible now. Let me know if there is anything that shouldn't go in or anything we need to discuss and I can fix it. Otherwise I think I am at a state I am fine merging this.

jrood-nrel:
I still don't really know what to do with `host_name` in the `darwin.yaml` env, but I will hopefully fix it when using the environment_updater.

psakievich (approved):

PR #6, opened 2021-08-19
Solve OpenFAST issue with incorrect HDF5 library names propagating to Nalu-Wind

jrood-nrel:


jrood-nrel:
Actually, I need to test this on Linux. It seems HDF5 installed with autotools has different compiler wrappers that get installed than an HDF5 installed with CMake. So I need to make sure this is a general solution. This is very confusing.

psakievich (approved):
This all looks good. Thanks for tracking all this down with hdf5.

jrood-nrel:
Tested this on linux and it works.

PR #7, opened 2021-08-23
Add rhodes configuration

jrood-nrel:


jrood-nrel:
Looks like this is working for me.

PR #14, opened 2021-09-07
Add eagle configuration

jrood-nrel:


jrood-nrel:
I think this is ready. Mostly building everything for me on Eagle now. I just need to update the exawind-driver to deal with cuda outside of this.

repos/exawind/packages/exawind/package.py
```diff
@@ -66,3 +66,8 @@ def copy_compile_commands(self):
             target = os.path.join(self.stage.source_path, "compile_commands.json")
             source = os.path.join(self.build_directory, "compile_commands.json")
             copyfile(source, target)
+
```
psakievich: Do you want to make this a variant?  If we are always going to add all the submodules then we should just use `submodules=True` in the version definition.

repos/exawind/packages/amr-wind/package.py
```diff
@@ -24,7 +24,8 @@ def cmake_args(self):
         saved_golds = os.path.join(os.getenv('SPACK_MANAGER'), 'tmp', 'tmp_golds', 'amr-wind')
         current_golds = os.path.join(os.getenv('SPACK_MANAGER'), 'golds', 'current', 'amr-wind')
         linked_golds = os.path.join(self.stage.source_path, "test", "AMR-WindGoldFiles")
-        options.append(define('AMR_WIND_SAVE_GOLDS', True))
+        # Make this a variant in the future
```
psakievich, line 27: Yeah we need to rework this. I'm getting install errors with AMR-wind due to these paths.

psakievich (commented):
Looks good.  Not sure about the Exawind submodules strategy though. See my comments

repos/exawind/packages/exawind/package.py
jrood-nrel: I think this might be a future task. Currently the exawind driver always enables testing, so we might as well set the submodules to true for now.

repos/exawind/packages/amr-wind/package.py
jrood-nrel: I think I just want to set this to false for now and rework it in another PR.

psakievich: yeah agreed

psakievich (approved):

PR #17, opened 2021-09-08
Improve error handling for find_machine

psakievich:
I've also made it so it's not an error in `start.sh` to be on a new machine that isn't established.


jrood-nrel (approved):
Looks really nice compared to the old logic.

PR #27, opened 2021-10-13
Add custom hypre parameters

jrood-nrel:


jrood-nrel:
It looks like does works as expected so I'm going to merge it. We can add on our own custom hypre configure parameters with this.

PR #31, opened 2021-10-21
Initial configuration for building Exawind for the CPU on Spock

jrood-nrel:


repos/exawind/packages/nalu-wind/package.py
```diff
@@ -10,9 +10,9 @@ class NaluWind(bNaluWind):
     variant('asan', default=False,
             description='turn on address sanitizer')
 
-    if os.environ['SPACK_MANAGER_MACHINE'] != 'summit':
-        depends_on('ninja', type='build')
-        generator = 'Ninja'
+    #if os.environ['SPACK_MANAGER_MACHINE'] != 'summit':
```
psakievich: no more ninja? or we could add something like 
```python
if os.environ['SPACK_MANAGER_MACHINE'] not in ['summit', 'spock']:
    ...
```

scripts/install-exawind.sh
```diff
@@ -24,4 +28,4 @@ cmd "spack env create -d ."
 cmd "cp ${SPACK_MANAGER}/env-templates/${SPACK_MANAGER_MACHINE}.yaml ${SPACK_MANAGER}/environments/exawind/spack.yaml"
```
psakievich: Would you being willing to try the `create_machine_spack_environment.py` script here instead?  It's doing the same thing but it is more general, and it is what the CI is using to generate the environments. I think it would be good to try merging our processes in this script as a starting point.

repos/exawind/packages/nalu-wind/package.py
jrood-nrel: I think we should only use ninja for trilinos. I don't think it's worth the trouble anywhere else.

scripts/install-exawind.sh
jrood-nrel: Yeah I can start trying that here.

psakievich: Also with that script you will be able to get rid of the template `spack.yaml` and just add
`create_machine_spack_environment.py -s exawind+hypre -d ${SPACK_MANAGER}/environments/exawind`

repos/exawind/packages/nalu-wind/package.py
psakievich: so you're not having trouble with ninja+trilinos on spock?

jrood-nrel: I haven't gotten that far. I'm currently at trilinos failing on missing blas routines.

repos/exawind/packages/trilinos/package.py
```diff
@@ -3,6 +3,3 @@
 
 class Trilinos(bTrilinos):
     patch('c++17.patch', when='cxxstd=17')
```
psakievich: We can actually remove this patch now. Trilinos merged my PR

scripts/install-exawind.sh
```diff
@@ -20,6 +20,10 @@ if [ "${SPACK_MANAGER_MACHINE}" == 'summit' ]; then
   cmd "module unload xl"
   cmd "module load python/3.7.7"
   cmd "module load binutils"
+elif [ "${SPACK_MANAGER_MACHINE}" == 'spock' ]; then
+  cmd "module unload rocm"
+  cmd "module load rocm/4.3.0"
```
psakievich: Does this work without manually loading the rcom module?  The `packages.yaml` you added looks like it should be loading it automatically

psakievich (commented):
Question and observation

repos/exawind/packages/trilinos/package.py
psakievich: Although c++17 wouldn't work with the master branch, but that is okay I suppose

psakievich:
Ugh did I push a style error?

jrood-nrel:
Yeah I think I will let you handle the style error.

psakievich (approved):

PR #34, opened 2021-10-26
Use MPI wrappers for TIOGA on Darwin.

jrood-nrel:
Also remove ninja for everything but Trilinos.

### DIFF
--- a/repos/exawind/packages/amr-wind/package.py
+++ b/repos/exawind/packages/amr-wind/package.py
@@ -4,8 +4,6 @@ import os
 from shutil import copyfile
 
 class AmrWind(bAmrWind):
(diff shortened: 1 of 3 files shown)


psakievich:
Should we just get rid of ninja for Trilinos too?  Is it worth keeping around if we have to fight with it?  It seems like an all or nothing kind of thing to me.

jrood-nrel:
It actually seems to work for trilinos and it gives us the most benefit, where it has constantly burned me in other packages. I don't see it as all or nothing.

psakievich:
Okay I'll follow your lead.

psakievich (commented):
LGTM

PR #35, opened 2021-10-26
Add mixed CPU/GPU exawind-driver capability for Eagle and Summit

jrood-nrel:
This also:
 1. Generalizes the darwin `compilers.yaml`
 2. Updates the `install-exawind.sh` script to use `create_machine_spack_environment.py`
 3. Moves the stage directory by default to a spack-manager stage directory
 4. Moves the cache directory by default to a spack-manager cache directory
 5. Updates the Spack submodule
 6. Includes the specific newer `ld` required on Eagle in `compilers.yaml`


jrood-nrel:
@psakievich Do you want to update the script to avoid blank yaml files? Then I can just have Eagle build clingo and we can keep the concretizer as clingo.

scripts/install-exawind.sh
```diff
@@ -9,19 +12,34 @@ set -e
 
 MYPWD=${PWD}
 
+cmd "export TMPDIR=${MYPWD}/../tmp"
+cmd "export SPACK_MANAGER=${MYPWD}/.."
+cmd "source ${SPACK_MANAGER}/start.sh"
+
 if [ "${SPACK_MANAGER_MACHINE}" == 'summit' ]; then
   cmd "module unload xl"
   cmd "module load python/3.7.7"
```
psakievich, line 21: Just a side note, but have you tried putting these in the `compilers.yaml` for eagle?

psakievich:
> @psakievich Do you want to update the script to avoid blank yaml files? Then I can just have Eagle build clingo and we can keep the concretizer as clingo.

@jrood-nrel yeah i'll update that tonight.

scripts/install-exawind.sh
jrood-nrel: I guess I should put binutils in there.

jrood-nrel: I added this.

psakievich (approved):

PR #39, opened 2021-10-28
Add custom M4

jrood-nrel:
Also need `nalu-wind+hypre` in the gpu-cpu mixed specs for Eagle and Summit.

### DIFF
--- a/repos/exawind/packages/m4/package.py
+++ b/repos/exawind/packages/m4/package.py
@@ -3,5 +3,5 @@ from spack.pkg.builtin.m4 import M4 as bM4
 import os
 
 class M4(bM4):
-    if os.environ['SPACK_MANAGER_MACHINE'] != 'summit':
+    if os.environ['SPACK_MANAGER_MACHINE'] == 'summit':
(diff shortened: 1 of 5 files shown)


repos/exawind/packages/m4/package.py
```diff
@@ -0,0 +1,7 @@
+from spack import *
+from spack.pkg.builtin.m4 import M4 as bM4
+import os
+
+class M4(bM4):
+    if os.environ['SPACK_MANAGER_MACHINE'] != 'summit':
```
psakievich:
```suggestion
    if os.environ['SPACK_MANAGER_MACHINE'] == 'summit':
```
Don't you mean?

psakievich (changes requested):

psakievich (approved):

repos/exawind/packages/m4/package.py
jrood-nrel: Oops. Didn't realize I could have just committed your changes. I even typed my passcode into Summit to fix it.

PR #42, opened 2021-10-30
Begin migration of scripting to be a spack extension

### LINKED ISSUE #38: Migrate scripts to be extensions of spack
We'd like to have all of our scripts become [extensions](https://spack.readthedocs.io/en/latest/extensions.html) of spack so the syntax would change to something like:
`spack manager create-env` 
`spack manager find-machine`
etc.

This will consolidate unit testing and formatting and bring us closer to contributing back to spack.

psakievich:
This is the first major step in converting to use [spack extensions](https://spack.readthedocs.io/en/latest/extensions.html) instead of just one off scripts (closes #38).
The benefits of this move are:

1.  Ability to truely extend spack instead of standard scripting
2. Ability to add spack internals into our unit tests
3. Further leverage of the spack framework (style guide, unit tests, etc)
4. A more common interface for users (all commands will be spack commands). Hopefully this makes spack less intimidating over time 🤞 

To separate `spack-manager` created commands from the native `spack` commands all `spack-manager` commands will be in a subparser and are accessed through `spack manager`. To see what the current commands are you can run `spack manager --help` and you can do the same on each of the functions within that subparser.

I'm interested to hear any thoughts or requests especially from the friendly beta testers @PaulMullowney, @neilmatula,  @ndevelder and @alanw0


jrood-nrel (approved):
This looks like a fantastic way to go.

jrood-nrel:
Heh, actually do we mean to update the submodule here? We actually want the latest Spack because it has more custom packages I upstreamed.

psakievich:
No I keep doing that on accident. I'm trying to fix a bug in spack too. So I keep jumping around on branches. I'll fix that.  I'm working on a better `develop` command right now. Hope to have that fixed this morning and then I'll get it all cleaned up.

psakievich:
@jrood-nrel right now I've left in all the old scripts so I don't break peoples work flow. I'm thinking follow on PR to delete them after we do some test driving

PR #61, opened 2021-12-07
Remove sierra-python dependencies

psakievich:
Sierra deleted the python3 module which breaks any build that depends on python for these machines.  Let's just build it ourselves from now on. With `spack concretize --reuse` we should really only have to build python once if we need it.

I think the only thing in our stack that relied on it (right now) is using ninja to build packages.

### DIFF
--- a/configs/ascicgpu/packages.yaml
+++ b/configs/ascicgpu/packages.yaml
@@ -22,12 +22,3 @@ packages:
     buildable: false
   cuda:
     version: [11.2.1]
-  python:
-    version: [3.6.10]
-    buildable: false
-    externals:
-    - spec: python@2.7.10
-      prefix: /projects/sierra/linux_rh7/install/Python/2.7/
-    - spec: python@3.6.10
-      modules:
-      - sierra-python/3.6.10
--- a/configs/cee/packages.yaml
+++ b/configs/cee/packages.yaml
@@ -20,15 +20,6 @@ packages:
   hdf5:
     version: [1.10.6]
     variants: +mpi+cxx+hl
-  python:
-    version: [3.6.10]
-    buildable: false
-    externals:
-    - spec: python@2.7.10
-      prefix: /projects/sierra/linux_rh7/install/Python/2.7/
-    - spec: python@3.6.10
-      modules:
-      - sierra-python/3.6.10
   zlib:
     version: [1.2.5]
     buildable: false


jrood-nrel:
Fine with me. Is the submodule update intended?

psakievich:
no it is not.  whoops

tasmith4 (approved):
This is reasonable to me in principle.  Do we have a bootstrapping problem though?  i.e. it requires Python3 even to run the spack manager commands (before we get to the builds), correct?

psakievich:
@tasmith4 yes that is correct for systems that don't have python3 natively installed. However right now that is just ascicgpu machines.  All the others have it native.  So I think we ask for it to get installed. In the mean time I created a build of python3 in the projects directory for those who don't have access to a module with python3 we can trust (like myself).

psakievich:
I've been thinking about how to deal with this sort of thing going forward. I'm increasingly feeling like we can't really rely on sierra modules to give us what we want for this project. It is more work than I'd like to keep up with their changes, and so it doesn't seem like a robust long term solution. I'm leaning toward creating my own modules more an more

tasmith4:
I mean, Exawind is its own project, so it's reasonable for it to have its own modules/build system/centrally managed TPLs/etc.  I think the biggest hangup has been someone needs to manage that and there's no dedicated Exawind DevOps support that I know of (everyone who does DevOps-type stuff is taking time out of their other work on the project).  So I think a lot of Sandians have defaulted to Sierra modules because they are familiar.  I agree that it's better in abstract to have our own stuff, but is it worth the long-term maintenance cost?  Maybe the answer is yes if Sierra is changing too often or doesn't provide what we really need?

psakievich:
yeah that is the hang up. However, I think most of these things we'd need are a one time install for the most part. that and via spack-manager I can create a script to repopulate the modules we need for a fresh install on a new system as well.  @jrood-nrel what are your thoughts on this? Do you think we could create a uniform admin environment that would install all the basic things we need (cmake, mpi, maybe compilers?) that could be stored as an environment template? seems like this would be a nice feature for making things uniform and help with your new modules issue we discussed

jrood-nrel:
I decided to do this very early on in the Exawind project at NREL. Then it became me installing all my own modules because I consider our default modules insufficient. This is how I do it https://github.com/jrood-nrel/spack-configs . I haven't had to update the modules in over a year, but you will definitely have build problems installing your own modules. I always have a large amount of them when I refresh the modules. I keep having the thought of how to do all that in spack-manager, but I haven't been able to understand how to automate it yet. I like the separation with the upstreams so it still feels like a step by step process with manual intervention like in my `install-modules.sh` script.

PR #66, opened 2021-12-09
Wrapper packages

tasmith4:
Fix what seems to be a typo in Nalu-Wind's wrapper package.

Add a wrapper package for Trilinos so we can turn on the STK unit tests.

I haven't written a lot of spack packages so if there are ways to clean this up please suggest.


psakievich (approved):
Looks great!

PR #67, opened 2021-12-09
Add saving of nalu golds

jrood-nrel:


repos/exawind/packages/nalu-wind-nightly/package.py
```diff
@@ -28,8 +28,6 @@ class NaluWindNightly(bNaluWind, CudaPackage):
     variant('host_name', default='default')
     variant('extra_name', default='default')
 
-    version('master', branch='master', submodules=True)
```
psakievich, line 31: I think we want to leave this in.  This ensures that the mesh submodule gets cloned too.

psakievich (commented):
LGTM, but I have one question

repos/exawind/packages/nalu-wind-nightly/package.py
jrood-nrel: I looked into this as I have been using my own repo here. I think this already does it https://github.com/psakievich/spack-manager/blob/c967ff4c4421c5b3c0414e775c63c2185bff7e84/repos/exawind/packages/nalu-wind/package.py#L8 or should we keep both?

psakievich: ah I see.  No that is fine then.  I thought we'd configured the nalu-wind package to only clone the mesh repo when we use a develop spec.  I guess that would be my preference since I will use this for generic installs too, but that doesn't need to be addressed in this PR.

psakievich: but for this package we definitely always want to clone the submodules.

psakievich (approved):

repos/exawind/packages/nalu-wind-nightly/package.py
jrood-nrel: That's true. I can leave it in. I vaguely remember trying to do submodules with a develop spec.

PR #74, opened 2021-12-14
Update some things in the test scripts

jrood-nrel:


jrood-nrel:
This PR doesn't do a whole lot, but it has some changes I need in the meantime while I think about how to optimize the cloning in the tests.

PR #75, opened 2021-12-16
Optimize cloning and updating of persistent repos in testing

jrood-nrel:
Trying something like this at the moment.


scripts/run-exawind-nightly-tests.sh
```diff
@@ -72,22 +83,31 @@ cmd "spack env activate ${EXAWIND_ENV_DIR}"
...
 cmd "tar -czf ${SPACK_MANAGER}/golds/archived/nalu-wind/nalu-wind-golds-${DATE}.tar.gz -C ${SPACK_MANAGER}/golds/tmp/nalu-wind ."
 
-printf "\nDone at $(date)"
+STAGE_DIR=$(spack location -S)
+if [ ! -z "${STAGE_DIR}" ]; then
+  #Haven't been able to find another robust way to rm with exclude
+  printf "\nRemoving all unused staged directories...\n"
+  cmd "cd ${STAGE_DIR} && rm -rf resource* spack-stage-b* spack-stage-c* spack-stage-d* spack-stage-f* spack-stage-g* spack-stage-hd* spack-stage-hw* spack-stage-i* spack-stage-j* spack-stage-k* spack-stage-l* spack-stage-m* spack-stage-nc* spack-stage-net* spack-stage-o* spack-stage-p* spack-stage-q* spack-stage-r* spack-stage-s* spack-stage-tar* spack-stage-ti* spack-stage-u* spack-stage-v* spack-stage-w* spack-stage-x* spack-stage-y* spack-stage-z*"
+  #Would like something like this
+  #find ${STAGE_DIR}/ -maxdepth 0 -type d -not -name "spack-stage-trilinos*" -exec rm -r {} \;
```
psakievich, line 110: I think we can do this with the spack python api.  This would be a good thing for us to pair on. There has to be a better way to do this.

psakievich: oh what if we just manually set the stage directory for trilinos and hypre instead?  `spack stage -p $SPACK_MANAGER/stage trilinos hypre` then you can just wipe this directory out if you want

jrood-nrel: Set it per spec? I think that might bring us back to the issue of needing something like hashes.

jrood-nrel:
Alright I'm going to merge this so I can try it out on the official test repo.

PR #78, opened 2021-12-16
Add in build line identifiers for trilinos version and amrex version

jrood-nrel:
Still need to test this.


jrood-nrel:
I think this is working as expected.

psakievich (approved):

PR #83, opened 2021-12-18
Fix path creation errors for amr-wind and nalu-wind

psakievich:


jrood-nrel:
Did you mean to update the submodule here?

psakievich:
No good catch.  Maybe I can make an action to not allow merging with a spack module update unless there is a label for it

jrood-nrel (approved):

PR #84, opened 2021-12-20
Create workflow for mapping views to externals

### LINKED ISSUE #62: Create way for users to use externals in dev process

psakievich:
Working on a way to map installs from spack views to externals that users can utilize in their own builds.
I've been hemming and hawing back and forth on this for a while, but I think it is best to break this up in to two parts.  

1) when a view is created, a file `externals.yaml` can be generated in the view directory that will contain the specs
for all the usable externals that are stored in the view.  This could be something we potentially push to spack at some point

2) when a developer wants to use this view they will call a command like `spack manager externals /path/to/view` which will search for the `externals.yaml` and copy it to the environment if it exists (plus add it to the `spack.yaml` includes)

I've got 1) pretty close to done, but need to work on 2). I also want to add some functionality to remove specs from the `externals.yaml` but I'm not sure if I should add it to step 1) or 2). Leaning toward 2) to give the users control.

Closes #62 when complete

Update:  I ended up making this a single step process. To use this the command should be pointed at a spack environment with a view.

`spack manager external $SPACK_MANAGER/environments/exawind/snapshots/{compiler}/{date} --blacklist {packages you want to develop}`


psakievich:
🤔 I just realized that for the default symlink views there is a directory that gets created automatically `view/.spack/[package]` that has all the build logs and a json description of the spec.  This is different than how we've been doing things (without projections), but perhaps it would make sense to use the projections.  I think the hard-link will do this as well. Parsing this data in step 2 would eliminate step 1 entirely.

But I guess it doesn't matter too much when this gets investigated since the end users workflow wouldn't change either way.

psakievich:
Well I can't get the unit test framework figured out right now, but this is working really well when I actually use it.  I want to just merge it and circle back on some Friday where I need a distraction

jrood-nrel (approved):

PR #94, opened 2021-12-22
Arcticus config

jrood-nrel:
This also updates Spack. I think we probably want https://github.com/spack/spack/pull/27846 . I realize this will have an effect on the tests when we update, but I guess it should also help test the latest Spack.


scripts/install-exawind.sh
```diff
@@ -34,15 +37,17 @@ else
   cmd "spack manager create-env -s exawind+hypre -d ${SPACK_MANAGER}/environments/exawind"
 fi
 
-cmd "cd ${SPACK_MANAGER}/environments/exawind"
-cmd "spack env activate -d ."
+printf "\nActivating Spack environment...\n"
+cmd "spack env activate -d ${SPACK_MANAGER}/environments/exawind"
 
 if [ "${SPACK_MANAGER_MACHINE}" == 'darwin' ]; then
```
psakievich, line 40: We actually don't need to call this. If compilers are missing the environment will call this for us

configs/arcticus/env.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash -l
+
+module use /soft/modulefiles
+module load intel/oneapi/release/2021.4.0
```
psakievich, line 4: Did you have to build these or are they on the system?

psakievich (approved):
Looking good.

scripts/install-exawind.sh
jrood-nrel: Oh interesting. I think I will leave it explicit for now, but add a comment.

configs/arcticus/env.sh
jrood-nrel: These are on the machine.

PR #110, opened 2022-01-20
View fix

psakievich:
Clean up some issues with the snapshot creator


jrood-nrel:
I don't think we want the spack submodule updated.

jrood-nrel (approved):

psakievich:
@jrood-nrel the first one was a mistake, but moving to the latest spack develop commit is not really very many changes so I say we do i

jrood-nrel:
Yeah that's fine.

PR #121, opened 2022-01-26
Add example developer env script

jrood-nrel:


psakievich:
This looks good.  There is nothing wrong with it but I would personally I would rather have this be a python script.  I feel like getting people to script with the `spack-python` shebang is going to be more beneficial in the long run.  It's the gateway drug to more powerful scripting and better parsing.  It would require users to have run `start.sh` or have run `spack-start` first, but I feel like that is a minimal ask.

jrood-nrel:
While I agree on doing everything in python in the long run would probably be better. The bash scripts like these are currently just my notes, so I prefer to store them here at the moment.

psakievich:
yeah that's fine by me.

psakievich (approved):

PR #138, opened 2022-02-01
Add auto assignment of env variable for external detections

psakievich:
Will need to update these files for eagle and summit

### DIFF
--- a/scripts/supported_external_paths.py
+++ b/scripts/supported_external_paths.py
@@ -1,0 +1,17 @@
+#!/usr/bin/env spack-python
+import manager_cmds.find_machine as find_machine
+
+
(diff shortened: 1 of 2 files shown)


jrood-nrel:
https://github.com/psakievich/spack-manager/blob/6e046b8bd0717ffc4ac24e32e9d8518d69445617/scripts/create-developer-env.sh#L26

jrood-nrel (approved):

psakievich:
@jrood-nrel so I didn't create this feature to rely on the symlink. it needs the location of your spack-manager installation on the machine, and then resolves the paths from there

jrood-nrel:
Ok I can update them after this is merged

PR #147, opened 2022-02-04
Use mostly external packages on Summit

jrood-nrel:
In an effort to get the exawind driver running on Summit, this is my latest attempt. Making it a draft until my job with the binaries from this runs.


psakievich (approved):
Nice!

jrood-nrel:
This builds and runs exawind if we want to use it. It didn't solve the nans I see in belos though for the turbine run.

psakievich:
I think we should test it out. It will make snapshots much smaller and build times trivial for most people.  If we have problems we can always switch back to building everything ourselves

PR #150, opened 2022-02-06
ASCICGPU: Config update

### LINKED ISSUE #148: Build is broken on ascicgpu22
While working with @ldh4 yesterday on ascicgpu22, we observed the following:
* Check out latest spack-manager (probable SHA ed0d592)-- got link errors relating to libgfortran (we interpreted to mean the compiler environment was messed up)
* Check out spack-manager SHA d08300b -- got same link errors
* Remembered you need to do `git submodule update` to get the spack version associated with that spack-manager SHA -- builds fine

Strangely, I was unable to reproduce on ascicgpu057.  From these facts I suspect a regression bug in spack sometime between those two spack-manager SHAs, that somehow only affects ascicgpu22.

Spec is `nalu-wind+cuda cuda_arch=70 ^trilinos+cuda+stk_unit_tests`.

psakievich:
Update ascicgpu to use compilers matching eagle. These compilers and externals were built with spack using the environment described by [env-templates/system_externals.yaml](https://github.com/psakievich/spack-manager/blob/main/env-templates/system_externals.yaml).

Closes #148

### DIFF
--- a/env-templates/system_externals.yaml
+++ b/env-templates/system_externals.yaml
@@ -4,25 +4,21 @@ spack:
   - system_specs:
     - cmake
     - mpich%gcc@9.3.0
+
   - when: env['SPACK_MANAGER_MACHINE']=='ascicgpu'
     system_specs: [cuda@11.2.2%gcc@9.3.0]
-
-  - when: env['SPACK_MANAGER_MACHINE']=='darwin'
-    system_specs: [gcc@9.3.0]
-
-  - when: env['SPACK_MANAGER_MACHINE']!='darwin'
-    system_specs: [gcc@9.3.0+binutils+piclibs]
-
   specs:
   - $system_specs
+  - binutils
+  - cuda@11.2.2
   view:
     default:
       root: /projects/wind/spack-manager/views/system
       projections:
-        all: '{compiler.name}-{compiler.version/}{name}-{version}'
+        all: '{compiler.name}-{compiler.version}/{name}-{version}'
       link_type: hardlink
   config:
-    install_missing_compilers: true
+    concretizer: original
   packages:
     all:
       target: [x86_64]
--- a/tests/ci_config_concretization.py
+++ b/tests/ci_config_concretization.py
@@ -25,18 +25,20 @@ def check_config_concretizations(name, ref_yaml):
 
 def run_tests(args):
     failure = False
-    machine_names = list(machine_list.keys())
-    matrix_test_machines = ['eagle', 'summit']
     this_machine = find_machine(verbose=False)
-
-    if '_matrix.yaml' in args.yaml:
-        machine_names = matrix_test_machines
+    # only test darwin on matching ci platform
+    if this_machine == 'darwin':
+        machine_names = ['darwin']
     else:
-        machine_names = list(set(machine_names) - set(matrix_test_machines))
-    # remove from ascicgpu from darwin due to gcc conflict
-    # need to test if we can move to boost@1.77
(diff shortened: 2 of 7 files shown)


psakievich:
This failure is because on darwin boost up to version 1.76 conflicts with gcc due to a bad system detection. I can either move darwin's boost up to 1.77 or disable darwin/ascicgpu combo from running

env-templates/system_externals.yaml
```diff
@@ -4,25 +4,21 @@ spack:
...
   view:
     default:
       root: /projects/wind/spack-manager/views/system
       projections:
-        all: '{compiler.name}-{compiler.version/}{name}-{version}'
+        all: '{compiler.name}-{compiler.version}/{name}-{version}'
       link_type: hardlink
   config:
-    install_missing_compilers: true
+    concretizer: original
```
tasmith4, line 21: what does "concretizer: original" mean?  are we not going to use clingo any more?  (don't really care, but curious)

tests/ci_config_concretization.py
```diff
@@ -26,11 +27,16 @@ def run_tests(args):
...
+    this_machine = find_machine(verbose=False)
 
     if '_matrix.yaml' in args.yaml:
         machine_names = matrix_test_machines
     else:
         machine_names = list(set(machine_names) - set(matrix_test_machines))
+    # remove from ascicgpu from darwin due to gcc conflict
+    # need to test if we can move to boost@1.77
+    if this_machine == 'darwin' and 'ascicgpu' in machine_names:
+        machine_names.remove('ascicgpu')
```
tasmith4, line 42: maybe I'm a little late to the party, but I'm confused by this issue.  I thought darwin was Mac?  Why does that have anything to do with ascicgpu?

env-templates/system_externals.yaml
psakievich: No this is a typo.  I need to update this. This template doesn't really affect how anything runs. It's just documentation of how I built the compiler, mpi library, and cuda that we will be using on ascicgpu machines

tests/ci_config_concretization.py
psakievich: I was going to try bootstrapping this set of compilers/builds onto my mac.

psakievich: Oh wait wrong file here. So @tasmith4 in our CI process we concretize each machine's config on two platforms (darwin and linux).  It's never caused a hang up before, but here since we're explicitly saying the compiler is gcc@9.7 it is causing a failure for darwin that I outlined above.  The way to solve this is to move boost up, or ignore this CI combo.  I went with the later since it doesn't require any more testing. 

Really we probably don't need to concretize all the systems on both platforms though.

tasmith4: Ok I see what you're saying.  I guess if it was up to me, since we know what platform each machine is ahead of time, I'd rather only concretize the correct platform for the specific machine.  I don't love the idea of introducing workarounds for the fact that we are attempting to concretize for mismatched machine/platform.

psakievich: sure I can make that change. this is mainly an artifact of standing up CI quickly.  The other idea was if we have platform independent concretization that is another win, but darwin has so many exceptions it may not be worth doing

psakievich: @tasmith4 update made to the CI

tasmith4: thanks!

tasmith4 (approved):

PR #158, opened 2022-02-08
Add documentation generation

jrood-nrel:


jrood-nrel:
Looks like this generated docs (with warnings I can fix later). We won't know if it works until we merge it though.

psakievich (approved):

PR #162, opened 2022-02-09
Add first text for snapshot workflow example

jrood-nrel:


jrood-nrel:
https://jrood-nrel.github.io/spack-manager has this pushed already so you can see the sphinx rendering.

docs/snapshot_workflow.md
```diff
@@ -0,0 +1,304 @@
...
+remote: Total 354065 (delta 0), reused 0 (delta 0), pack-reused 354065
+Receiving objects: 100% (354065/354065), 155.14 MiB | 22.08 MiB/s, done.
+Resolving deltas: 100% (150589/150589), done.
+Submodule path 'spack': checked out '3576e5f3d6b34d8bc8c8c8f2749127ece1ce89be'
+```
+
+We `cd` to `spack-manager` and activate it:
+```
+[user@el1 spack-manager]$ export SPACK_MANAGER=${SCRATCH}/test/spack-manager && source ${SPACK_MANAGER}/start.sh
+Activating Spack-Manager...
```
psakievich: For this section I'd prefer to show people to source the `useful_bash_functions.sh` and have them use `spack-start` instead.  I think that is the model we want to consistently show since that will also enable the `quick-*` commands as well.  

I'd really like for everyone to activate spack-manager that way since it enables the reduced syntax commands.  What are your thoughts?

docs/snapshot_workflow.md
```diff
@@ -0,0 +1,304 @@
...
+enabled (`+hypre`), and CUDA (`+cuda cuda_arch=70`), using the GCC compiler (`%gcc`, which without a version, selects the default compiler version).
+```
+[user@el1 spack-manager]$ spack manager create-env --name exawind --spec 'nalu-wind@master+hypre+cuda cuda_arch=70 %gcc'
+making /scratch/user/test/spack-manager/environments/exawind
+```
+
+Once the environment is created, we need to activate it:
+```
+[user@el1 spack-manager]$ spack env activate -d ${SPACK_MANAGER}/environments/exawind
+```
```
psakievich: here we could add 
```
Both of these steps can be combined into 1 with `quick-create-env -n exawind -s "nalu-wind@master+hypre+cuda cuda_arch=70%gcc"`
```

docs/snapshot_workflow.md
```diff
@@ -0,0 +1,304 @@
...
+
+Next, we decide how to take advantage of the prebuilt snapshots on the machine. Here we use the `spack manager external`
+command to specify a "view" in which we want to pull external packages into our environment. Snapshots are organized by date.
+On Eagle there are also symlinks for the latest view available. We will use the latest snapshot, and one that is attributed to 
+our GCC with CUDA configuration, called `gcc-cuda`. Views are typically organized by compiler, e.g. `intel`, `clang`, `gcc`, and
+`gcc-cuda`, etc. We also need to "blacklist" any packages we plan on developing locally. Even without building source locally in
```
psakievich: If we add develop specs first and then call external it will auto black-list them for us.  It would be good to add that into the documentation here too.

docs/snapshot_workflow.md
```diff
@@ -0,0 +1,304 @@
...
+* update_golds_10_26_2021
+```
+
+Once our externals and git clones are configured, we have the necessary `*.yaml` files in our `${SPACK_MANAGER}/environments/exawind` environment directory
+to "concretize" and (re)install our entire project. The `spack.yaml` file in this directory is the main yaml file in which the
+other yaml files are included. Concretizing is required to solve or map our loosely defined `nalu-wind@master+hypre+cuda cuda_arch=70 %gcc` spec into "concrete" parameters of our dependency graph (or DAG). The concrete DAG is _exactly_ how Spack will fulfill the dependencies for your spec. We
+concretize with the command (we almost always want to use the force with `-f`):
+```
+[user@el1 spack-manager]$ spack concretize -f
+==> Warning: the original concretizer is currently being used.
```
psakievich, line 124: we should add a comment about this warning in the documentation explaining what it is

psakievich: Also users don't _have_ to concretize here anymore since `spack install` does it. It might be worth taking this section out and having a section on when to reconcretize with the `-f` flag?

psakievich (commented):
Great progress. Added some thoughts

docs/snapshot_workflow.md
jrood-nrel: Sure, I can do that.

docs/snapshot_workflow.md
jrood-nrel: Sure

docs/snapshot_workflow.md
jrood-nrel: Should that be the `start.sh` then? `useful_bash_functions.sh` is pretty long and doesn't indicate the entry point.

psakievich: hmm good point

docs/snapshot_workflow.md
jrood-nrel: I kind of think concretization should be an explicit step because it's a main idea within Spack.

docs/snapshot_workflow.md
psakievich: The reason for keeping them separate is so people don't have to load spack into their shell until they want it

jrood-nrel: Is there a case where you would want spack-manager activated and not the spack within it? I can't say that is something I would ever do.

psakievich: Well spack-manager isn't activated. Just the scripts to activate it when you want are defined in your shell. To do that you just have SPACK_MANAGER defined and the bash scripts sourced. Then when you call any of the quick-commands it activates spack and spack-manager as part of the process. The idea is minimally invasive additions to the shell to make it activatable at any point.

jrood-nrel: That's fair, but I think everything should go into `start.sh` then like I've done here. I don't think useful_bash_functions should be an entry point.

start.sh
```diff
@@ -1,39 +1,178 @@
...
+  if [[ "${SPACK_MANAGER_MACHINE}" == "NOT-FOUND" ]]; then
+      echo "Machine not found."
+  fi
+  if [[ -n "$(${SPACK_MANAGER}/scripts/supported_external_paths.py)" ]]; then
+      export SPACK_MANAGER_EXTERNAL=$(${SPACK_MANAGER}/scripts/supported_external_paths.py)
+  fi
+  export PATH=${PATH}:${SPACK_MANAGER}/scripts
+}
+
+# function to quickly activate an environment
```
psakievich, line 49: why not just `source ${SPACK_MANAGER}/scripts/useful_bash_functions.sh`?

psakievich (changes requested):
I like the idea of including `useful_bash_functions.sh` into `start.sh` but I think this change loses the intent behind the `quick-*` commands. See my comment in #170

start.sh
jrood-nrel: Because if `useful_bash_functions` is the intended entry point, I really think entry should begin in `start.sh`. We can move the `spack-start` to a file in `scripts` if it matters.

psakievich: yeah this fine. I'm on board

start.sh
```diff
@@ -1,39 +1,178 @@
...
+  fi
+  
+  ########################################################
+  # Environment stuff
+  ########################################################
+  export SPACK_ROOT=${SPACK_MANAGER}/spack
+  export SPACK_DISABLE_LOCAL_CONFIG=true
+  export SPACK_USER_CACHE_PATH=${SPACK_MANAGER}/.cache
+  export PYTHONPATH=${PYTHONPATH}:${SPACK_MANAGER}/scripts:${SPACK_MANAGER}/spack-scripting/scripting/cmd:${SPACK_MANAGER}/spack-scripting/scripting
+  source ${SPACK_ROOT}/share/spack/setup-env.sh
```
psakievich:
```suggestion
if [ -z ${SPACK_MANAGER_MACHINE} ]; then
  export PYTHONPATH=${PYTHONPATH}:${SPACK_MANAGER}/scripts:${SPACK_MANAGER}/spack-scripting/scripting/cmd:${SPACK_MANAGER}/spack-scripting/scripting
  source ${SPACK_ROOT}/share/spack/setup-env.sh
 fi
```

Maybe something like this to make `spack-start` a no-op if it has already been called?

start.sh
```diff
@@ -1,39 +1,178 @@
...
-export PATH=${PATH}:${SPACK_MANAGER}/scripts
+#!/bin/bash
+
+cmd() {
+  echo "+ $@"
+  eval "$@"
+}
+
+# function to initialize spack-manager's spack instance
+function spack-start() {
```
psakievich, line 22:
```suggestion
function spack-manager-start() {
```

jrood-nrel: So I was going to make this change, but I think I would decide against it now. I think we should consider `source start.sh` as having spack-manager activated, but `spack-start` as having activated spack. It's also shorter.

psakievich: okay that sounds good to me

psakievich (approved):
LGTM. Let's make a post about this on slack to let the user base know, and add the `remove-spack-prompt` function added in #169. I'll post on slack

psakievich:
closes #163

PR #167, opened 2022-02-10
Catch error when users don't give a path for external

### LINKED ISSUE #166: spack manager external without --latest gives confusing error
`TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType`

psakievich:
closes #166 .

The issue is since the path is no longer required it can be None and still proceed.  This change ensures that it is not None before moving on.  Improved the error message, and made it so the concretizer doesn't change unless we actually create the external.yaml without errors.


jrood-nrel (approved):

psakievich:
@jrood-nrel do you think we should invert the behavior of this command? make `--latest` be the default behavior, and have users supply `--path` instead?

jrood-nrel:
Sure, that makes sense.

PR #168, opened 2022-02-10
Make execution match help for create-dev-env

### LINKED ISSUE #164: spack manager create-dev-env documentation error
When running `spack manager create-dev-env` with no args, it fails, asking for a spec, but the help text says the spec is optional.  Running with spack manager checked out at 9c2fd5e (note that in the help text it has square brackets around all options, implying they are optional):

```
$ quick-develop
+ spack-start
==> Removing cached information on repositories
+ spack manager create-dev-env
==> Error:
ERROR: specs are a required argument for spack manager create-dev-env.


ERROR: Exiting quick-develop prematurely

$ quick-develop --help
+ spack-start
==> Removing cached information on repositories
*************************************************************
HELP MESSAGE:
quick-develop sets up a developer environment and installs it

This command is designed to require minimal arguments and simple setup
with the caveat of accepting all the defaults for:

- repo and branch cloned for your packages
- latest external snapshot with the default compilers/configurations

Please note that for specifying multiple specs with spaces you need to
wrap them in quotes as follows:

"'amr-wind@main build_type=Debug' nalu-wind@master 'exawind@master build_type=Debug'"

The base command and it's help are echoed below:


+ spack manager create-dev-env --help
usage: spack manager create-dev-env [-h] [-m MACHINE] [-d DIRECTORY | -n NAME]
                                    [-y YAML] [-s SPEC [SPEC ...]]

optional arguments:
  -d DIRECTORY, --directory DIRECTORY
                        Directory to copy files
  -h, --help            show this help message and exit
  -m MACHINE, --machine MACHINE
                        Machine to match configs
  -n NAME, --name NAME  Name of directory to copy files that will be in $SPACK_MANAGER/environments
  -s SPEC [SPEC ...], --spec SPEC [SPEC ...]
                        Specs to populate the environment with
  -y YAML, --yaml YAML  Reference spack.yaml to copy to directory
*************************************************************
```

psakievich:
As outlined in #164 `spack manager create-dev-env` is not consistent with the `--help` prompt.  This makes it so it is not an error to omit specs or to give non-concrete specs. However, non-concrete specs won't be made into develop specs and automatically cloned.  A warning message will be printed for each of these specs.

Closes #164

### DIFF
--- a/spack-scripting/scripting/cmd/manager_cmds/create_dev_env.py
+++ b/spack-scripting/scripting/cmd/manager_cmds/create_dev_env.py
@@ -3,6 +3,7 @@ import argparse
 import manager_cmds.create_env as create_env
 import manager_cmds.develop
 
+import spack.cmd
 import spack.environment as ev
 
 
@@ -21,9 +22,10 @@ def create_dev_env(parser, args):
     env_path = create_env.create_env(parser, args)
     env = ev.Environment(env_path)
     ev.activate(env)
-    for s in env.user_specs:
+    specs = spack.cmd.parse_specs(args.spec)
+    for s in specs:
         # check that all specs were concrete
-        if not s.concrete:
+        if not s.versions.concrete:
             print('\nWarning: {spec} is not concrete and will not '
                   'be setup as a develop spec.'
                   '\nAll specs must be concrete for '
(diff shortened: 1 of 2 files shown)


tasmith4:
@psakievich so what will happen if I don't give any specs?

psakievich:
@tasmith4 nothing. You won't get any warnings, and you won't get any specs. It will be an empty, but machine configured environment. The behavior should be the exact same as `spack manager create-env` since you aren't giving any new information to act on.

psakievich:
I'm working on a couple of unit tests for this right now

tasmith4:
That makes sense then, because I could still come back to add develop specs.

psakievich:
yeah you only get the new behavior if you meet the minimum criteria for that behavior, otherwise we fall back to the parent function.  I think this is the most logical path forward.

psakievich:
@tasmith4 unit tests added to confirm the new behavior

tasmith4 (approved):

PR #169, opened 2022-02-10
Attempt for removing the prompt

### LINKED ISSUE #165: quick-* commands add text to prompt
Is it possible for the `quick-*` commands to not add text to the prompt by default?  I've discovered it doesn't play well with my custom prompt.

psakievich:
Addresses #165 

@tasmith4 ~~I can't get the function to do what you want. The alias is dangerous because if the `$SPACK_OLD_PS1` is not set then it will just wipe the prompt totally.  Might be best to go the environment variable route. @jrood-nrel is more of a bash wizard then me. Maybe he has an idea...~~

Or I could just double check my bash syntax...

This is working now.  @tasmith4 let me know if you have suggestions for an alternate name for the function

### DIFF
--- a/scripts/useful_bash_functions.sh
+++ b/scripts/useful_bash_functions.sh
@@ -140,7 +140,7 @@ The base command and it's help are echoed below:
 }
 
 # function to remove spack prompt from the shell
-function remove_spack_prompt() {
+function remove-spack-prompt() {
   if [[ "$1" == "-h" || "$1" == "--help" ]]; then
     echo "This command removes a spack added shell prompt (if it exists) that signifies the current environment name."
   fi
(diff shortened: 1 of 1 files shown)


psakievich:
closes #165

scripts/useful_bash_functions.sh
```diff
@@ -138,3 +138,14 @@ The base command and it's help are echoed below:
   fi
   cmd "spack install"
 }
+
+# function to remove spack prompt from the shell
+function remove_spack_prompt() {
```
tasmith4: The name is fine, but it would be easier to remember if it used "-" not "_" as a separator, for consistency with the other functions in this file

tasmith4 (changes requested):

scripts/useful_bash_functions.sh
psakievich: ah good point

tasmith4 (approved):

PR #170, opened 2022-02-10
Perform shell checks on useful_bash_functions.sh

### LINKED ISSUE #163: Specifying SPACK_MANAGER in environment
I was trying to use the new reduced workflow commands as follows, starting from empty directory `spack-manager-test`, in a fresh shell:
```
spack-manager-test$ git clone --recursive git@github.com:psakievich/spack-manager
# a bunch of git output, spack-manager is checked out at 9c2fd5e
spack-manager-test$ source spack-manager/scripts/useful_bash_functions.sh
spack-manager-test$ quick-develop
+ spack-start
-bash: /start.sh: No such file or directory

ERROR: Exiting quick-develop prematurely
```

This error goes away if I run `export SPACK_MANAGER=$(pwd)/spack-manager` -- is the fact that the useful bash functions fail without that environment variable a workflow feature or bug?

If it's considered a feature, I'd like to request all these functions fail immediately if SPACK_MANAGER is unset with a clear message asking the user to set it.

psakievich:
This performs the same shell checks on the `useful_bash_functions.sh` as we use in `start.sh` due to the corner case outlined in #163 .  It would be nice if we could just auto set `$SPACK_MANAGER` using the location of the script.  I started looking into this rabbit hole, but it seems like there are too many corner cases: 

https://stackoverflow.com/questions/59895/how-can-i-get-the-source-directory-of-a-bash-script-from-within-the-script-itsel
https://stackoverflow.com/questions/24112727/relative-paths-based-on-file-location-instead-of-current-working-directory

I think it is safest to just put this burden on the users.  I also made these errors return instead of exit since this is a sourced script so it would kill people's login shell before they can see the error message. (Infinite loop of frustration).

Closes #163

### DIFF
--- a/scripts/useful_bash_functions.sh
+++ b/scripts/useful_bash_functions.sh
@@ -1,5 +1,18 @@
 #!/bin/bash
 
+########################################################
+# Tests same as start.sh for now
+########################################################
+if [[ -z ${SPACK_MANAGER} ]]; then
+  echo "ERROR: Env variable SPACK_MANAGER not set. You must set this variable."
+  return 1
+fi
+
+if [[ ! -x $(which python3) ]]; then
+  echo "WARNING: spack-manager is only designed to work with python 3."
+  echo "You may use spack, but spack-manager specific commands will fail."
+fi
+
 cmd() {
   echo "+ $@"
   eval "$@"
--- a/start.sh
+++ b/start.sh
@@ -5,7 +5,7 @@
 ########################################################
 if [[ -z ${SPACK_MANAGER} ]]; then
     echo "Env variable SPACK_MANAGER not set. You must set this variable."
-    exit 125
+    return 1
 fi
 
 if [[ ! -x $(which python3) ]]; then


scripts/useful_bash_functions.sh
```diff
@@ -1,5 +1,18 @@
...
 
+########################################################
+# Tests same as start.sh for now
+########################################################
+if [[ -z ${SPACK_MANAGER} ]]; then
+  echo "ERROR: Env variable SPACK_MANAGER not set. You must set this variable."
+  return 1
+fi
+
+if [[ ! -x $(which python3) ]]; then
```
tasmith4, line 11: if users have to source start.sh before running any spack-manager-related python commands, I think we don't need this check here.  All the functions in this file that make spack manager calls source start.sh first, right?

tasmith4:
@psakievich agreed auto-set would be nice, but it does seem like there's too many corner cases.  I think printing a clear error message is good enough, since it clues a user in immediately to what they need to do.

jrood-nrel:
I think `useful_bash_functions.sh` should be merged into `start.sh` as I have done in #162 .

psakievich:
@jrood-nrel The main idea behind the `quick-*` commands is that they are a one command to activate/set things up.  #162 makes it so there are a minimum of two commands.  `source start.sh` and `spack-start`.  And now using the `quick-*` commands takes three commands, `source start.sh`, `spack-start`, and `quick-*`.   I think this is the wrong direction since @tasmith4 want's as few commands as possible.

I agree we should include `useful_bash_functions.sh` into `start.sh`, but I also think we should make it so `spack-start` and/or `source start.sh` are no-op's on the environment and leave `spack-start` as the first call in the commands to preserve this behavior.

I think we can do this by:

1.  Checking for `$SPACK_ROOT` and not sourcing spack if it is defined
2. Adding a check so we don't keep appending `$PYTHONPATH` (probably need another environment variable) 

I think those are the only dangers of sourcing `start.sh` multiple times.  The big slow down is sourcing spack.

jrood-nrel:
I put`spack-start` back into each of the quick commands, or you can call `spack-start` directly. I think this is the same behavior we currently have if not using `start.sh` directly, which I think has only been myself. So it's just my workflow that changes, but I may be missing something.

jrood-nrel (approved):

psakievich:
closing due to changes in #162

PR #184, opened 2022-02-14
Docs, and build-env-dive

psakievich:
I'm adding some documentation and another convenience script for diving into the build environment, given a spec.  @tasmith4 I'm especially interested in your input on the latter since I think this has been a requested feature from the stk team.

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -191,7 +191,7 @@ function build-env-dive() {
     echo "This command will move to the build directory for the spec provided, and then dive into the build environment in a new subshell."
   fi
   if [[ -z ${SPACK_ENV} ]]; then
-    echo "You must have an active environment to use sm-dive."
+    echo "You must have an active environment to use build-env-dive."
     return 1
   fi
   cmd "spack cd -b $*"
--- a/start.sh
+++ b/start.sh
@@ -9,15 +9,15 @@ cmd() {
 # Tests
 ########################################################
 if [[ -z ${SPACK_MANAGER} ]]; then
-    echo "Env variable SPACK_MANAGER not set. You must set this variable."
-    exit 125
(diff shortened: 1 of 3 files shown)


jrood-nrel (approved):

start.sh
```diff
@@ -185,3 +185,18 @@ function remove-spack-prompt() {
     unset SPACK_OLD_PS1
   fi
 }
+
+# function for diving into the build environment from a spack build
+function sm-dive() {
```
tasmith4: So, in Sierra World, the acronym "SM" means something else ... could we consider another name for this?

tasmith4 (changes requested):
other than the name, I think this looks great!

start.sh
psakievich: any suggestions?  build-env-dive?

tasmith4: @psakievich I actually like build-env-dive a lot

psakievich: okay `build-env-dev` it shall be

start.sh
```diff
@@ -185,3 +184,17 @@ function remove-spack-prompt() {
...
   fi
 }
+
+# function for diving into the build environment from a spack build
+function build-env-dive() {
+  if [[ "$1" == "-h" || "$1" == "--help" ]]; then
+    echo "This command will move to the build directory for the spec provided, and then dive into the build environment in a new subshell."
+  fi
+  if [[ -z ${SPACK_ENV} ]]; then
+    echo "You must have an active environment to use sm-dive."
```
tasmith4: also needs a name change here

tasmith4 (changes requested):

tasmith4 (approved):

PR #191, opened 2022-02-17
Add a useful links section to the front page of the docs

tasmith4:
Right now just has the developer quick start.  I put it above benefits since most people would care more about useful links than benefits on the second and succeeding times they visit the website.


psakievich:
@tasmith4 I added a table of contents tree where you wanted the useful links.  Does this also meet your intent? I like the headers you've added though

tasmith4:
Yeah I think what you did also works good

PR #195, opened 2022-02-17
Split out spack-start into its own script

tasmith4:
Move the internals of the `spack-start` convenience function to another script that can be sourced by users who want to setup spack/spack manager correctly, but don't want to source the bash-specific convenience functions.


psakievich (approved):
LGTM

PR #198, opened 2022-02-17
Add some more intro documents

psakievich:


psakievich:
Merging due to how late it is, and that it's all docs

PR #207, opened 2022-02-22
Add hypre-mini-app

jrood-nrel:


psakievich (approved):
LGTM. are there any cmake flags about tests that we should be aware of or are they always on since it's a mini app?

jrood-nrel:
No tests. It just loads a matrix in the same manner as Nalu-Wind.

PR #212, opened 2022-02-23
Add configuration for crusher

jrood-nrel:


psakievich (approved):
I think we should look into improving our concretization CI to test the HIP builds since that is maturing and more users are building with HIP. Doesn't have to be in this PR.

PR #213, opened 2022-02-24
use actual hostnames for SNL machines

tasmith4:
It seems like the full machine name in MachineData is only used for text output, so we can just use the actual hostname, at least for SNL machines.

@psakievich or @jrood-nrel please speak up if I'm missing something and this won't work.

### DIFF
--- a/spack-scripting/scripting/cmd/manager_cmds/find_machine.py
+++ b/spack-scripting/scripting/cmd/manager_cmds/find_machine.py
@@ -41,11 +41,11 @@ match
 machine_list = {
     # SNL
     'cee': MachineData(lambda: is_cee(socket.gethostname()),
-                       'cee.snl.gov'),
+                       socket.gethostname()),
     'snl-hpc': MachineData(lambda: is_snl_hpc(socket.gethostname()),
-                           'snl-hpc.snl.gov'),
+                           socket.gethostname()),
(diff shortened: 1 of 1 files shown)


jrood-nrel:
We shouldn't use a facility to get the hostname in my opinion. It should be listed as we want it to appear and expect it to appear. Although it may work for your machines, people will copy and paste this and it won't work in general. For example here is `hostname` on Crusher:
```
$ hostname
login2
```

tasmith4:
So part of the motivation is to distinguish between machines ... e.g. `cee-compute004.sandia.gov` and `ews00904.sandia.gov` are both part of the CEE group, but even though the same software configs work for both, the two machines are actually quite different hardware-wise.  Even `ascicgpu22.sandia.gov` and `ascicgpu057.sandia.gov` have differences that have mattered to spack-manager, even though they are pretty close.

How would we distinguish between machines like this without using a facility?

jrood-nrel:
It would be better to use something other than `hostname` to create a full host name. For the non-SNL machines we use environment variables that the computing facility has set. However, I don't know the Sandia machines, so if it works best for you and Phil, it's fine with me. Just be aware it may not act like you expect in several situations.

psakievich (approved):
This should be fine for SNL machines. I think we are consistent there.

jrood-nrel (approved):

PR #217, opened 2022-03-01
Posix compliant type check on `_spack_start_called` function

marchdf:
According to [this](https://unix.stackexchange.com/questions/332005/test-for-functions-existence-that-can-work-on-both-bash-and-zsh), `type -t` is not posix compliant. And indeed it breaks on `zsh` shells. The link proposes a fix and it is implemented in this PR. I tested both `zsh` and `bash` and it seems to work.

### DIFF
--- a/scripts/spack_start.sh
+++ b/scripts/spack_start.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-if [[ $(type -t _spack_start_called) != function ]]; then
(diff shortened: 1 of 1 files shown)


psakievich (approved):
I have done no testing with zsh. We may need to consider adding it if it is a common shell for users.

marchdf:
It's the default shell on OSX so in that regard it's probably worth testing for it... Thanks for merging!

PR #218, opened 2022-03-01
Use generic mtune/march on Summit

jrood-nrel:


jrood-nrel:
I don't think this is the right approach.

PR #220, opened 2022-03-02
Turn off STK SIMD

jrood-nrel:


psakievich:
@jrood-nrel this looks good. I think we should make this a variant and default it so stk SIMD is turned off.  I would presume that the STK team will be interested in addressing this at some point and it would make their life easier.  Plus it would be good to be able to flip the switch off and on periodically for testing.

tasmith4:
@jrood-nrel what’s the motivation for turning off SIMD by default project-wide? I thought it was only an issue on Crusher?

jrood-nrel:
It's a problem on Summit as well. SIMD has caused segfaults on every machine I've used, most notably non-Intel machines. It's recently caused numerous segfaults in the unit tests on an Intel machine with GCC. I don't have any doubt it should be disabled in general.

psakievich:
We saw this on the ascicgpu machines too. I turned SIMD off by setting the arch to x86_64.

tasmith4:
@jrood-nrel @psakievich ok, that seems reasonable, I had no idea this was that bad.  Is there a github issue posted, and is Alan aware of this?

I think we should discuss this at the next Thursday meeting, can one of you bring it up if I forget?  I'm curious how much people even care about having SIMD at all -- should we be considering hardcoding it off in code rather than adding an option to the package?

psakievich:
Yeah those are good points. Let's bring it up at the next meeting.

jrood-nrel:
Aha, I can finally run the turbine on both Crusher and Summit with this. Finding the general target for Summit removed the power9 flags from the command line and stops the nans in Belos.

psakievich (approved):

jrood-nrel:
Ok I'm going to merge this. We can talk about other approaches to this in the future.

PR #222, opened 2022-03-02
Add capability for using hypre in mixed device regime

jrood-nrel:
This isn't fun, but necessary to have this working.


jrood-nrel:
It actually seems like this is a bad idea linking to the same library twice.

PR #230, opened 2022-03-09
Add nvcc-wrapper in nalu-wind to OpenMPI and MPT

jrood-nrel:
Let's treat all the MPIs the same in `nalu-wind`: https://github.com/psakievich/spack-manager/pull/229


psakievich (approved):
Yeah this is a good idea.

jrood-nrel:
It's possible we might want to do this for HIP, but I will wait on that.

PR #232, opened 2022-03-10
Add full-repo option

psakievich:
This PR will address @PaulMullowney's requests for cloning

### DIFF
--- a/spack-scripting/scripting/cmd/manager_cmds/develop.py
+++ b/spack-scripting/scripting/cmd/manager_cmds/develop.py
@@ -21,6 +21,14 @@ def git_clone(branch, repo, path, shallow, all_branches):
     git(*git_args)
 
 
+def git_remote_add(path, name, repo):
+    """
+    git remote add wrapper
+    """
+    git = spack.util.executable.which('git')
+    git('-C', path, 'remote', 'add', name, repo)
+
+
 def _redundant_code_from_spack_develop(args):
     """
     Re-use the path and spec checking from spack.cmd.develop
@@ -81,6 +89,12 @@ def manager_develop(parser, args):
 
     spack_develop.develop(None, args)
 
+    if args.add_remote:
+        # a clone must have taken place at this point so we can
+        # safely add a repo
+        remote_name, remote_repo = args.add_remote
+        git_remote_add(path, remote_name, remote_repo)
+
 
 def add_command(parser, command_dict):
     subparser = parser.add_parser('develop', help='a more intuitieve interface for '
@@ -95,5 +109,8 @@ def add_command(parser, command_dict):
     subparser.add_argument('--all-branches', '-ab', required=False,
                            action='set_true', help='clone all branches '
                            'of the repo', default=False)
+    subparser.add_argument('--add-remote', nargs=2, metavar=('remote_name',
+                           'remote_repo'), required=False,
+                           help='add a remote as part of the clone')
 
     command_dict['develop'] = manager_develop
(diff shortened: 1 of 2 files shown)


spack-scripting/scripting/cmd/manager_cmds/develop.py
```diff
@@ -90,5 +106,11 @@ def add_command(parser, command_dict):
                              help='git repo to clone from')
     subparser.add_argument('--shallow', required=False, action='store_true',
                            help='performa a shallow clone of the repo')
+    subparser.add_argument('--all-branches', '-ab', required=False,
```
tasmith4: Looks like `-ab` is intended as a short arg here?  I think this will be too confusing, since standard unix command line short options are only one char, and `-ab` would typically be equivalent to `-a -b`.

spack-scripting/scripting/cmd/manager_cmds/develop.py
```diff
@@ -7,18 +7,28 @@
 from spack.error import SpackError
 
 
-def git_clone(branch, repo, path, shallow):
+def git_clone(branch, repo, path, shallow, all_branches):
```
tasmith4, line 10: If I was giving a real "I see myself as a code owner" review, I'd complain here because I strongly dislike boolean args (because they require that a function do more than one thing, and because the combinatorial space that needs to be tested gets large very quickly as you start adding them).  But I'll let you decide how seriously you want to take that feedback ... ;)

spack-scripting/tests/test_develop.py
```diff
@@ -33,4 +33,5 @@ def test_requireUserToSpecifyGitRepoAndBranch(self, mock_git_clone, mock_develop
             path = os.path.join(e.path, name)
             manager('develop', '--repo-branch', repo, branch, spec)
             mock_develop.assert_called_once()
-            mock_git_clone.assert_called_once_with(branch, repo, path, False)
+            mock_git_clone.assert_called_once_with(
```
tasmith4, line 53: looks like we are not testing all possible combinations for the new arg?

tasmith4 (changes requested):

spack-scripting/scripting/cmd/manager_cmds/develop.py
psakievich: sure. I'll just remove the short arg. I don't think we need it.  We violate this a lot, and I hadn't thought about that until now.  I'm hesitant to change the api, but we should add that to a list of things to do in a major refactor.

spack-scripting/scripting/cmd/manager_cmds/develop.py
psakievich: It's a valid critique.  It's a violation of the single purpose principle. I'm glad you brought it to my attention, but I'd prefer to not worry about it here.  This is a polymorphic function since it is wrapping git clone, but i'm only supporting a few options and there are only two booleans.   If it grows anymore then I will refactor.

spack-scripting/tests/test_develop.py
psakievich: no not yet. I haven't added any new tests in this PR

psakievich: I'll add some more tests today.

spack-scripting/scripting/cmd/manager_cmds/develop.py
psakievich: tests added. these tests just assert that the flags lead to the correct git calls.  I globed them into one parameterized test that will test every combination of args.

spack-scripting/tests/test_develop.py
tasmith4: looks good

tasmith4 (approved):
Looks good @psakievich, thanks!

PR #241, opened 2022-03-12
Further work on externals and snapshots

psakievich:
PR to fix some issues

- [ ]  Make projection path change if cuda/rocm is turned on rather than if it is just a dependency. Our cuda and non-cuda binaries were being assigned to the same directory in the projections (explains behavior on eagle) (Actually not sure if this is an issue. it is not on ascicgpu after all with the changes below)
- [x] Make all the specs get written properly to the `externals.yaml` file

### DIFF
--- a/scripts/snapshot_creator.py
+++ b/scripts/snapshot_creator.py
@@ -150,9 +150,9 @@ def add_view(env, extension, link_type):
         'root': view_path,
         'projections': {'all': '{compiler.name}-{compiler.version}/{name}/'
                         '{version}',
-                        '^cuda': '{compiler.name}-{compiler.version}-'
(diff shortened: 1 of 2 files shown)


psakievich:
I really hope this is the final nail in the coffin on this externals stuff.  The file generation process is much better.  We should test it out on more machines.

psakievich:
I've tested on ascicgpu. cuda and non cuda builds from the same externals.yaml file.  Now just multi-compiler on eagle and then I think we can be confident. It would be good to figure out how to unit test this stuff but setting up a fabricated environment in the unit tests we can pull externals from is too hard for me right now.

jrood-nrel (approved):

PR #251, opened 2022-03-22
Split out exawind-tests-script.sh

tasmith4:
I'm assuming this script was written on the fly for historical reasons, but now it is completely generic with respect to machine, and is even written into the `$SPACK_MANAGER/scripts` directory, so it seems easier to maintain if it is a standalone script that always lives in `scripts` instead of being written from another script every time.


jrood-nrel:
Maybe it's just semantics, and this is how I did it in the past, but I would prefer not to have two separate test scripts.

psakievich:
I think I agree with @tasmith4 here. It seems strange to be cat'ing a whole script into our scripts directory every night when the script is not ever changing.  It also seems like a logical partition. The new file is stand alone, machine independent, and run-able script and what is lift is simply to control the submission details on different platforms.

jrood-nrel:
Yeah that is how I did it for years, but I specifically made it this way so everything necessary for testing is now in one script and it's much shorter than in the past. I don't think you can run the generated test script without `SPACK_MANAGER` set, so I don't think it stands on its own. It's going to be hard to sell me on this. Maybe if the `scripts` directory had less scripts in it.

psakievich:
That's fine. I don't really feel very strongly about this. So if you do then I'm happy to defer to your choice.

PR #253, opened 2022-03-23
Added the SNL 'ascic' platform

tasmith4:
will be primarily used for CPU nightly testing at SNL


tasmith4:
@psakievich pushing this now because I want it to run with the nightly tests tonight.  I'm happy to do a new PR with any changes you think are appropriate.

PR #263, opened 2022-04-19
Add more CPU-only testing lines at SNL, remove the massive tolerances

tasmith4:
Removing the tolerances because the current gold file strategy is to create initial gold files from a run on this machine, so these unvetted gold files already match anyway.

@jrood-nrel maybe we can talk about the best way to handle the extra name.  On STK, we wanted test lines that didn't have all the extra packages, and we wanted at least one debug-mode build, but these use cases aren't currently supported by the auto-generated extra name in `nalu-wind-nightly/package.py`.  For now I just hard-coded some, but maybe there's a way to do this more cleanly in the package.py.

### DIFF
--- a/env-templates/exawind_ascic_tests.yaml
+++ b/env-templates/exawind_ascic_tests.yaml
@@ -4,4 +4,8 @@ spack:
   concretization: separately
   view: false
   specs:
-    - 'nalu-wind-nightly +fftw+tioga+hypre+openfast abs_tol=1e100 rel_tol=1e100 %gcc@9.3.0'
+    - 'nalu-wind-nightly +fftw+tioga+hypre+openfast %gcc@9.3.0 extra_name=-gcc@9.3.0+tioga+hypre+openfast'
+    - 'nalu-wind-nightly %gcc@9.3.0 extra_name=-gcc@9.3.0'
+    - 'nalu-wind-nightly %gcc@9.3.0 buildType=Debug extra_name=-gcc@9.3.0-Debug'
+    - 'nalu-wind-nightly %intel@20.0.2 extra_name=-intel@20.0.2'
+    - 'nalu-wind-nightly %clang@10.0.0 extra_name=-clang@10.0.0'


psakievich (approved):

psakievich:
so one thing about this is it is going to overload our cdash. we only get 8 lines with the free version.  We need to upgrade to the paid version

tasmith4:
@psakievich I didn't realize we had a limit.  How realistic is it to upgrade to paid and on what timeframe could we do that?  Alternatively, do we need to have a broader discussion about which lines to run to keep it to 8?

psakievich:
@tasmith4 we've discussed it in the past and I don't think it should be a big deal.  @jrood-nrel has more information there. He's looked into it in the past IIRC

jrood-nrel:
The administrative effort behind subscribing to cdash for more build lines is the true cost of paying for cdash. I would prefer the NREL tests have priority since they've been going the whole project. I also think it's supposed to be 10 builds for free, so we could experiment.

env-templates/exawind_ascic_tests.yaml
```diff
@@ -4,4 +4,8 @@ spack:
   concretization: separately
   view: false
   specs:
-    - 'nalu-wind-nightly +fftw+tioga+hypre+openfast abs_tol=1e100 rel_tol=1e100 %gcc@9.3.0'
+    - 'nalu-wind-nightly +fftw+tioga+hypre+openfast %gcc@9.3.0 extra_name=-gcc@9.3.0+tioga+hypre+openfast'
+    - 'nalu-wind-nightly %gcc@9.3.0 extra_name=-gcc@9.3.0'
+    - 'nalu-wind-nightly %gcc@9.3.0 buildType=Debug extra_name=-gcc@9.3.0-Debug'
+    - 'nalu-wind-nightly %intel@20.0.2 extra_name=-intel@20.0.2'
+    - 'nalu-wind-nightly %clang@10.0.0 extra_name=-clang@10.0.0'
```
jrood-nrel, line 11: These `extra_name` entries are redundant.

tasmith4:
Makes total sense for NREL tests to have priority, what I meant by a broader discussion is that there are a couple motivations for Sandia to care about having local testing:  1) We've historically had a hard time reproducing nightly results from NREL platforms, among other reasons due to limited access and platform-specific issues e.g. on eagle, and 2) We (at least STK) are interested in other testing lines with slightly different options (such as building without the extra openfast/hypre/tioga dependencies).  Motivation 1 would lead us to want to duplicate the NREL lines to expose any platform-specific issues between NREL/SNL and enable Sandia developers to easily reproduce failures that occur across all sites.  Motivation 2 would lead us to have different lines than NREL so we can see some configurations that don't show up in NREL's lines.  If we're capped at some number of lines, Motivations 1 and 2 are going to be in conflict and we'll have to resolve that somehow (if we can't get unlimited lines, presumably by assuming the 4 standard NREL lines are always there and having an internal Sandia discussion about which lines we want in the remaining space).

@jrood-nrel what happens if we submit more lines than we're limited to?  Does it crash the whole dashboard or do the latest lines just lose and get silently dropped?

Also @jrood-nrel who do we need to talk to about setting up a paid CDash plan to find out the feasibility of applying the administrative effort to subscribe?

jrood-nrel:
I think these should all fit. I don't think I would bother trying to pay for CDash just yet.

tasmith4:
@jrood-nrel we also had 6 total GPU lines in mind to run ... :)

tasmith4:
Closing for now since I'm rethinking some elements of the strategy

PR #266, opened 2022-04-26
Remove need for original concretizer with externals

psakievich:
Based on the diagnosis of https://github.com/spack/spack/issues/28201 we can get clingo to work for our external specs by stripping `patches=*` variants from the specs.


jrood-nrel (approved):
Excellent.

PR #272, opened 2022-05-05
Refactor extra-name creation and clean up duplication

tasmith4:
* Cleans up some duplication introduced by the "dashboard_name" parameter
* Makes it so variants are added to the "extra_name" automatically, only for SNL dashboard builds
* ~~Switch from using "default" as the default value for "host_name" and "extra_name" to the None singleton~~ (can't do this cause it makes Spack mad)
* Slight rearrangement so that the cuda version is listed with the compiler version, and the trilinos version comes last (reflecting standard spack spec arrangement)
* For Cuda builds, always give the Trilinos UVM setting as part of the build name

### DIFF
--- a/env-templates/exawind_ascic_tests.yaml
+++ b/env-templates/exawind_ascic_tests.yaml
@@ -4,6 +4,6 @@ spack:
   concretization: separately
   view: false
   specs:
-    - 'nalu-wind-nightly+snl +fftw+tioga+hypre+openfast %gcc@9.3.0 dashboard_name="-release+fftw+tioga+hypre+openfast"'
-    - 'nalu-wind-nightly+snl %gcc@9.3.0 dashboard_name="-release"'
-    - 'nalu-wind-nightly+snl %gcc@9.3.0 build_type=Debug dashboard_name="-debug"'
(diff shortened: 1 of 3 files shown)


tasmith4:
@jrood-nrel Although this will slightly change the format of the existing NREL build names, I don't expect it to add clutter, since it's basically a rearrangement, but let me know if there's a concern with this that I'm not thinking of.  I expect the existing build names to look like this in the new format:
```
Linux-gcc@9.3.0-cuda@11.2.2 ^trilinos@develop+uvm
Linux-clang@10.0.0 ^trilinos@develop
Linux-gcc@9.3.0 ^trilinos@develop
Linux-intel@20.0.2 ^trilinos@develop
```

jrood-nrel:
Let's not put spaces in the build names.

tasmith4:
@jrood-nrel I wondered about that -- what goes wrong if we have spaces?  I initially added them because it breaks it up and makes it easier to read.

jrood-nrel:
I don't think we want to find out.

tasmith4:
I don't know, we love finding out new things cause we're scientists! ;)  ... uh, I mean, I'll get rid of the white space.

psakievich (approved):

PR #277, opened 2022-05-11
Add hdf5 option to amr-wind

jrood-nrel:
This has a few major changes like updating Spack and changing the HDF5 version, but I think we should make them.

### DIFF
--- a/configs/base/packages.yaml
+++ b/configs/base/packages.yaml
@@ -3,7 +3,7 @@ packages:
     version: [1.76.0]
     variants: cxxstd=14
   hdf5:
-    version: [1.12.2]
+    version: [1.10.7]
     variants: +cxx+hl
   netcdf-c:
     version: [4.7.4]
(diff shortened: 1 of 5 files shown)


psakievich (commented):
I think these all look reasonable to me. Updating HDF5 will cause a lot of rebuilds though so we should probably make an announcement about it on the spack-manager slack channel.

jrood-nrel:
I'm still working through this because Spack builds HDF5 through CMake now which causes all kinds of issues. CMake's FindHDF5 is much better when HDF5 is built with autotools.

psakievich:
Closes #281

jrood-nrel:
Also, I want to note I didn't inherit here `h5z-zfp` because its build system changed to cmake on the master branch which is what we want here. I have https://github.com/spack/spack/pull/29952 open, but in my opinion the `package.py` here is what should be in Spack and the versions before cmake for `h5z-zfp` should be ignored.

psakievich:
Yeah this looks good. I feel like your package changes wont get merged until @alalazo finishes his work supporting multiple build systems.

PR #279, opened 2022-05-12
Add framework for testing nalu-wind-nightly package

tasmith4:
* For now, we don't concretize the spec, in order to make the test predictable and robust -- i.e. we don't want to be checking for specific variants, and then fail later when someone makes another variant default.
  * Not concretizing has the added advantage of not slowing the tests down by several seconds per spec.
  * The fixture does memoize packages, so we only have to create one for each unique abstract spec if we do decide we really need concretization.
* Also, this adds *.swp to the .gitignore

First step towards resolving #280

### DIFF
--- a/spack-scripting/tests/conftest.py
+++ b/spack-scripting/tests/conftest.py
@@ -7,3 +7,7 @@ from spack.test.conftest import *  # noqa: F401
 @pytest.fixture(scope='session')
 def exawind_repo_path():
     yield os.path.join(os.environ['SPACK_MANAGER'], 'repos', 'exawind')
+
+@pytest.fixture(scope='session')
+def builtin_repo_path():
+    yield os.path.join(os.environ['SPACK_MANAGER'], 'spack', 'var', 'spack', 'repos', 'builtin')
--- a/spack-scripting/tests/conftest.py
+++ b/spack-scripting/tests/conftest.py
@@ -7,6 +7,7 @@
 
 
 import os
+
 import pytest
 
 import spack.paths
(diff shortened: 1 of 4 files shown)


spack-scripting/tests/test_nalu_wind_nightly.py
```diff
@@ -0,0 +1,5 @@
+import pytest
+import importlib
+
+nalu_wind_nightly = importlib.import_module('...repos.exawind.packages.nalu-wind-nightly.package', 'package')
+#from ...repos.exawind.packages.nalu-wind-nightly.package import *
```
psakievich:
```suggestion
from spack.pkg.exawind.nalu_wind_nightly import NaluWindNighlty 
```
Have you tried this?

spack-scripting/tests/conftest.py
```diff
@@ -1 +1,34 @@
...
+import pytest
+
+import spack.repo
+from spack.spec import Spec
 from spack.test.conftest import *  # noqa: F401
+
+
+@pytest.fixture(scope='session')
+def builtin_repo_path():
+    yield os.path.join(os.environ['SPACK_MANAGER'], 'spack', 'var', 'spack', 'repos', 'builtin')
```
psakievich: This is already defined with `spack.paths.packages_path`

tasmith4: fixed, thanks

tasmith4:
@psakievich I thought it would be good to get this testing framework into the repo, independent of the other bug fix, so I'm marking as ready for review now.

psakievich (approved):
Looks great! Thanks for doing this

PR #283, opened 2022-05-18
Update snl-hpc configuration

lawrenceccheung:
I updated the `compile.yaml` and `packages.yaml` specifications for the `snl-hpc` machines.  I wasn't able to get intel compilers to work on the cluster, nor things to run with mpich.  However, I was able to use the gcc compilers and the prebuilt openmpi modules to get things to compile and run.   

These changes should allow things to work out of the box for skybridge/chama, using a spec like:  
```
spack manager create-env --name exawind_gcc -m snl-hpc --spec 'nalu-wind@master+openfast amr-wind@main+openfast %gcc@7.2.0'
```

Lawrence

### DIFF
--- a/configs/snl-hpc/packages.yaml
+++ b/configs/snl-hpc/packages.yaml
@@ -4,12 +4,12 @@ packages:
       read: group
       group: wg-sierra-users
     compiler:
+    - gcc@7.2.0
     - intel@19.0.3.199
-    - gcc@8.3.1
     providers:
       mpi:
-      - mpich
       - openmpi
+      - mpich
       blas:
       - netlib-lapack
       lapack:
@@ -28,6 +28,11 @@ packages:
     variants: +mlx5-dv+optimizations
   openmpi:
     variants: ~vt+pmi schedulers=slurm fabrics=auto legacylaunchers=true 
+    buildable: false
+    externals:
+    - spec: openmpi@3.1.6 %gcc@7.2.0
+      modules:
+      - cde/prod/gcc/7.2.0/openmpi/3.1.6
   slurm:
     externals:
     - spec: slurm@20.02.6


lawrenceccheung:
Tagging @ndevelder on this as well to see if his experiences line up with this PR.

ndevelder:
This isn't how I did it...not sure if we should ultimately rely on CDE (gcc-7.2.0) compilers/libs or not. I used devtoolset-8 like in the default, but changed configs on openmpi and removed mpich because a dep wouldn't build. Do we know of any benefits to using gcc-8.3.1 vs. gcc-7.2.0? I can't find anything in 8 release notes other than better diagnostic messages.

psakievich:
So gcc-7.2.0 doesn't support `std::filesystem` which the driver uses and this was the original motivation for moving to gcc8.  @jrood-nrel has since moved this to a soft dependency. But I am inclined to not really on GCC 7 if we don't have to. @ndevelder can you post what TPL was having issues with MPICH?

jrood-nrel:
I would suggest not using GCC 7.

ndevelder:
So it was UCX that errored on build...I think this is true for both me and Lawrence. 

Also I was getting mpi errors with your variant setup in openmpi, so I ended up specifying version 4.0.5 and only using "fabrics=auto schedulers=auto" as variants. I now get an xpmem warning, so my build might not be optimal, but it runs!

psakievich:
Okay for now let's just have you guys use what's working for you. I think we should coordinate closer to get a better fix.

lawrenceccheung:
Thanks @psakievich.  Yeah, Nate and I have something working for the immediate near term (like today/tomorrow), but we'll set up something with you to discuss making official, production releases of Exawind on snl clusters.

Lawrence

psakievich:
@lawrenceccheung @ndevelder I'm testing this build out with `gcc@10.3.0` using #293.

ndevelder:
One more consideration...I just built with the latest intel from the sierra project, using openfast 3.1, and saw cases that were projected to take 64 hours to finish (using a gcc build) cut down to 44 hours....it's only one data point, but definitely makes me think twice about sticking with gcc?

psakievich:
@ndevelder this new configuration also adds an intel compiler. I can test with that as well.

ndevelder:
@psakievich thanks for adding that! I wrote too quickly and didn't look at the file changes...sorry. That sounds great. I'm also happy to test the intel build if you're short on time.

psakievich:
Superseded by #293

PR #293, opened 2022-06-01
Update SNL configs to utilize CDE spack work

### LINKED ISSUE #256: Should we add upstreams to the machine configs?
https://github.com/spack/spack/pull/29680 makes it so we can add upstreams in our spack.yaml/include.yaml files.  So what do you guys think about adding upstreams as part of the standard confgis? Do we want to go that route or only use externals?
@tasmith4 @jrood-nrel

psakievich:
This adds upstreams for the work done on tlcc2 and cee machines by the CDE project.  With these changes we will be utilizing their compilers and builds directly through spack.

I also copied the compilers.yaml between all the cee based machines since they were all using the same compilers anyways (minus the 2019 intel compiler no one was using; I deleted that one here too )

Closes #256 
Supersedes #283

### DIFF
--- a/configs/ascic/compilers.yaml
+++ b/configs/ascic/compilers.yaml
@@ -15,13 +15,75 @@ compilers:
 - compiler:
     environment: {}
     extra_rpaths:
-    - /opt/rh/devtoolset-8/root/usr/lib64
-    modules: []
+    - /projects/sierra/linux_rh7/SDK/compilers/gcc/7.2.0-RHEL6/lib64
+    - /projects/sierra/linux_rh7/SDK/compilers/gcc/7.2.0-RHEL6/lib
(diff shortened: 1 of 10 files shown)


tasmith4 (approved):
Looks great!

jrood-nrel (approved):

psakievich:
I tested this on `skybridge` and `ascicgpu22`. 

I built `nalu-wind` and `amr-wind` with intel, clang and gcc on skybridge (clang with amr-wind failed due to an ascent dependency, but everything else built).

On `ascicgpu22` I only built `nalu-wind%gcc@10.3.0` but I also started building with intel and clang and they were both building TPL's just fine.

psakievich:
`nalu-wind+openfast+hypre` builds with all 3 compilers on `ascicgpu22`.  I'm going to merge this.

PR #299, opened 2022-06-07
Pin trilinos to commit from 6-5-2022

jrood-nrel:
I think it's time to start pinning trilinos like we talked about. We will update the commit when we need any significant changes for AMD GPU work.

### DIFF
--- a/repos/exawind/packages/trilinos/package.py
+++ b/repos/exawind/packages/trilinos/package.py
@@ -10,6 +10,9 @@ from spack.pkg.builtin.trilinos import Trilinos as bTrilinos
 import os
 
 class Trilinos(bTrilinos):
+
+    version('develop', commit='7498bcb9b0392c830b83787f3fb0c17079431f06') # develop@6-5-2022
+
     variant('stk_unit_tests', default=False,
             description='turn on STK unit tests')
     variant('stk_simd', default=False,


psakievich:
I think we should have two versions. `stable` and `develop` where `stable` become the default version `nalu-wind` uses.  I think it is important that we keep testing the bleeding edge of trilinos to look at whats coming, and also to allow stk/trilinos teams to work on the latest versions of the code more seamlessly.

tasmith4:
probably in this PR just rename "develop" -> "stable", then the latest develop will still get automatically picked up.

jrood-nrel:
I think spack-manager should be able to build trilinos successfully by default.

psakievich:
@jrood-nrel yes I agree with you. But I think we can just make `stable` become the default.  I think the default should build but we also should be able to build (and test) the bleeding edge.  This is really just a package thing in nalu-wind and the driver in determining what the default version of trilinos is.

tasmith4:
@jrood-nrel my specific concern is that overriding the develop version in spack-manager's trilinos package is likely to cause confusion.  Many people have a very specific expectation about what they will get when they put `trilinos@develop` in a spec, i.e. that they are building the tip of Trilinos' develop branch, and this PR as written would not meet that expectation.  I think it would cause a lot of confusion if `trilinos@develop` meant something different depending on whether you built with spack-manager or regular spack, and would make it difficult to diagnose issues users encounter.  I think if we instead named the new version `stable`, and made it the default version when building spack-manager's trilinos package, then we would successfully build Trilinos by default with spack-manager and nalu-wind, while also maintaining expectations about the meaning of version names in specs.

jrood-nrel:
Maybe it's not a lot of work to permeate a stable version of trilinos through spack-manager, but I don't think I want to be the one to attempt it or keep track of it so I think I will just close my PR.

PR #300, opened 2022-06-07
Update Crusher configuration

jrood-nrel:


jrood-nrel:
This works for me successfully on Crusher.

repos/exawind/packages/trilinos/package.py
```diff
@@ -10,6 +10,9 @@
 import os
 
 class Trilinos(bTrilinos):
+
```
psakievich:
```suggestion
```
Let's deal with this one in #299 😉

psakievich (changes requested):

repos/exawind/packages/trilinos/package.py
jrood-nrel: Oh yeah I forgot about that here.

psakievich (approved):

PR #301, opened 2022-06-07
Add 'stable' trilinos version

psakievich:
Continuing #299

### DIFF
--- a/configs/base/packages.yaml
+++ b/configs/base/packages.yaml
@@ -20,7 +20,7 @@ packages:
   yaml-cpp:
     version: [0.6.3]
   trilinos:
-    version: [develop]
+    version: [stable, develop]
     variants: ~adios2~alloptpkgs~amesos+amesos2~anasazi~aztec+belos+boost~chaco~complex~debug~dtk~epetra~epetraext+exodus+explicit_template_instantiation~float+fortran~fortrilinos+glm+gtest+hdf5~hypre~ifpack+ifpack2~intrepid~intrepid2~isorropia+kokkos~mesquite+metis~minitensor~ml+mpi+muelu~mumps~nox~openmp~phalanx~piro~python~rol~rythmos~sacado+shards~shylu+stk~stratimikos~suite-sparse~superlu~superlu-dist~teko~tempus+teuchos+tpetra+uvm~x11~xsdkflags+zlib+zoltan+zoltan2 gotype=long cxxstd=14 build_type=Release
   all:
     variants: build_type=Release +mpi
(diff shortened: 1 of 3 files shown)


psakievich:
depends on https://github.com/spack/spack/pull/31036

psakievich:
I've tested this and it will make the default version of trilinos become `stable` using the spack-manager environment creation process.  Just need to get the update to the base package in. We can decide what to do with the testing in a follow on PR.

tasmith4:
Won’t putting “stable” in the base nalu-wind package break builds if we don’t also put stable in the base Trilinos package?

tasmith4:
Do we even need or want a specific trilinos version as a dependency in the base nalu-wind version? Why not just say “trilinos+options we need” as the dependency, then use the explicit spec or defaults from the yaml files in the environment to determine the correct version of trilinos during concretization?

psakievich:
@tasmith4 the syntax there is stable and anything of higher precedent. So this includes master and develop and just omits the current releases which we know typically wont build. Is this okay with you?

tasmith4:
@psakievich oops, I missed the colon after "stable" in your spack PR, so that's why I thought this was locked into the stable version.  I'm still slightly fuzzy on what's actually excluded by that spec (i.e. I forget the precedence rules for versions in spack -- if you can link me to them in the docs that would be awesome).  I'm also a little unclear on why we want to specify any specific version in the nalu-wind base package.  Not super opposed, but I'd like to hear a concrete argument for why we should put it there instead of in spack-manager's yaml files.  More precisely, I want to know what are the general principles we are following for when to put things in spack packages vs yaml files, so I can see how those apply in this instance.

psakievich:
@tasmith4 so the logic behind these infinity versions is that they are typically associated with a branch while numbered versions correspond to a release.  The reason for having the constraint on specific versions in the package is it keeps users from using a release since we know nalu-wind won't work with any trilinos release for very long.

This is not so much an issue in spack-manager, but it is in vanilla spack where we set this constraint to keep potential users from trying to build with versions we know will break. Since we inherit from this package we are bound by the same constraints, and it just seems like best practice to be consistent.  I guess it is not necessary in this commit since we will be bound by this constraint either way. 

I'm not sure where these infinity versions exist in the documentation, but they are in the source code here:
https://github.com/spack/spack/blob/3eba28e3830012b30a9ff37040903b28b007098c/lib/spack/spack/version.py#L61-L62

Their ordering is also their precedent so `develop > master > stable` wihich also matches the ordering we should expect for our pinned commit.

tasmith4:
@psakievich thanks that makes sense.  I wanted to put the comment about "develop at 6-5-22" as an argument to `version`, so that it would show up in the `spack info` call, but a brief tour of the spack code and docs suggest this isn't possible.  There's a lot of indirection in the code though, so if I'm wrong and we can do that, I'd like it if we did.

tasmith4 (approved):

PR #304, opened 2022-06-13
Allow the gold files generated in nightly testing to be placed in an arbitrary directory

tasmith4:
The environment variable `SPACK_MANAGER_GOLDS` can now be exported prior to running the testing script.  If it's not defined, the testing script will default to the current `$SPACK_MANAGER/golds`.

### DIFF
--- a/scripts/run-exawind-nightly-tests.sh
+++ b/scripts/run-exawind-nightly-tests.sh
@@ -30,11 +30,11 @@ else
     printf "\nSPACK_MANAGER set to ${SPACK_MANAGER}\n"
 fi
 
-if [[ -z ${SPACK_MANAGER_GOLDS} ]]; then
-    printf "\nSPACK_MANAGER_GOLDS not set so setting it to ${SPACK_MANAGER}/golds\n"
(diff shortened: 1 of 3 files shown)


scripts/run-exawind-nightly-tests.sh
```diff
@@ -30,6 +30,13 @@ else
     printf "\nSPACK_MANAGER set to ${SPACK_MANAGER}\n"
 fi
 
+if [[ -z ${SPACK_MANAGER_GOLDS} ]]; then
```
jrood-nrel: Can you name it so it's clear the variable is a directory?

jrood-nrel (approved):

psakievich (approved):
LGTM. Can you make the same change to AMR-Wind's package too though so we remain consistent?

scripts/run-exawind-nightly-tests.sh
tasmith4: good point.  done in latest commit.

tasmith4:
@psakievich done

PR #308, opened 2022-06-14
Stop relying on defaults in SNL nightly specs

tasmith4:
For some unknown reason the old specs are no longer concretized correctly.  In any case, since we want to test these specific specs, not just the defaults, it makes sense to be more explicit.

### DIFF
--- a/env-templates/exawind_ascicgpu_tests.yaml
+++ b/env-templates/exawind_ascicgpu_tests.yaml
@@ -2,7 +2,7 @@ spack:
   include:
   - include.yaml
   concretizer:
-    unify: when_possible
+    unify: false
   view: false
   specs:
     - 'nalu-wind-nightly+snl +fftw+tioga+hypre+openfast+cuda cuda_arch=70 %gcc@9.3.0 build_type=Release ^trilinos@develop+uvm'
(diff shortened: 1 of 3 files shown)


scripts/run-exawind-nightly-tests.sh
```diff
@@ -92,7 +92,7 @@ printf "\nActivating environment...\n"
 cmd "spack env activate -d ${EXAWIND_ENV_DIR}"
 
 printf "\nConcretizing environment...\n"
-cmd "spack concretize -f"
+cmd "spack concretize -f --fresh"
```
psakievich: You may or may not want this. `--fresh` will also update low level dependencies when their spack hashes change. So every time we change things this may create larger rebuilds and more bloat in the size of the spack repo.

env-templates/exawind_ascicgpu_tests.yaml
```diff
@@ -5,9 +5,9 @@ spack:
     unify: when_possible
```
psakievich:
```suggestion
    unify: false
```
This actually may be your issue.  `unify:when_possible` makes an effort to reduce the DAG as much as possible. It is a beta feature so I may have added it too soon.

scripts/run-exawind-nightly-tests.sh
tasmith4: removed.

env-templates/exawind_ascicgpu_tests.yaml
tasmith4: working after changed to `unify: false`

psakievich (approved):

PR #312, opened 2022-07-12
install nightly testing packages in serial

tasmith4:
Turns out that the SNL GPU testing machines can't handle multiple tests running simultaneously.  #311 doesn't quite fix it, because unlike on eagle, we are building multiple variants of `nalu-wind-nightly` each time this script runs.  It seems that overlapping `spack install` commands allows spack to simultaneously install multiple `nalu-wind-nightly` packages, which can cause the tests to conflict with each other.

This PR removes the `spack install` loop, to force the test packages to be built in serial and guarantee they won't conflict with each other.

Related to Exawind/nalu-wind#994

### DIFF
--- a/scripts/run-exawind-nightly-tests.sh
+++ b/scripts/run-exawind-nightly-tests.sh
@@ -113,7 +113,7 @@ cmd "spack clean -s || true"
 
 printf "\nTests started at: $(date)\n\n"
 printf "spack install \n"
-time (for i in {1..4}; do spack install --keep-stage & done; wait)
+time (spack install --keep-stage)
 printf "\nTests ended at: $(date)\n"
 
 printf "\nSaving gold files...\n"
--- a/scripts/run-exawind-nightly-tests.sh
+++ b/scripts/run-exawind-nightly-tests.sh
@@ -113,7 +113,11 @@ cmd "spack clean -s || true"
 
 printf "\nTests started at: $(date)\n\n"
 printf "spack install \n"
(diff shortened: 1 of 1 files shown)


psakievich:
This is strange. Each of those should be getting their own resource pool for building and testing.  Is it that the tests are conflicting, or that the gpu is running out of memory? This also might be something we want to raise with spack

tasmith4:
I suspect it's the GPU running out of memory, but I can't claim to know for sure.  The main datapoint I have is that there are none of these failures when running a single spack install locally, but they do show up when using this loop.  Not sure how to narrow it down to a spack bug if there is even one there though.

In general I tend to like running in serial on the nightly test better anyway, since it eliminates one more thing that can go wrong or cloud the results.

psakievich (changes requested):
I think we should do a conditional for this based on if it is ascicgpu for now and put a comment to revisit this.  This will slow down the build/test process on every machine and we really just need it for the 1.

tasmith4:
@psakievich done

psakievich (approved):

jrood-nrel:
This isn't the right change for the NREL tests. I'm also not so sure this is the correct solution for the GPU tests. I think it's probably exposing another issue.

jrood-nrel:
Oh I guess there is a conditional on it now. I think that is fine with me. I still think running multiple tests on a single GPU is fine. It may depend on the mode your GPUs are set, if it's in an exclusive mode or not.

PR #316, opened 2022-07-15
Add netcdf option to openfast for FSI

psakievich:
netcdf was added to openfast as part of the FSI work.  We need this option. defaulting to false.


psakievich:
I'm just going to merge this if the CI passes.  I need it for tomorrow and I don't think reviewers will be available until next week.  We can revert later if necessary.

jrood-nrel:
Does this require turning it on in OpenFAST somehow?

psakievich:
In the FSI stuff there will be a PR for openfast.  We're working with that unmerged PR right now for testing

PR #319, opened 2022-07-20
Add a clean function for spack-manager

psakievich:
Tribal knowledge function for cleaning things up: `sm-clean`. Let me know if I'm missing anything. Pinging @PaulMullowney as well.

Also adding some missing help messages from functions.

### DIFF
--- a/docs/general/FAQ.md
+++ b/docs/general/FAQ.md
@@ -2,8 +2,9 @@
 
 
 ## I'm getting lots of weird Spack errors, what should I do?
-If you have another Spack instance you've used in the past conflicts can occur due to the `~/.spack` directory.
-The first step we recommend is deleting this directory and trying again.
+If you have multiple spack instances or have recently updated spack things can get a little mangled.
+Please try the `sm-clean` function (available after `source $SPACK_MANAGER/start.sh`)  and see if it solves your issue.
+If this doesn't work then raise/create an issue on slack/github and we will help you figure it out.
 
 ## What is Clingo and why is Spack saying it is __bootstrapping__ it?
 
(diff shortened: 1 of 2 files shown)


docs/general/FAQ.md
```diff
@@ -2,8 +2,9 @@
 
 
 ## I'm getting lots of weird Spack errors, what should I do?
-If you have another Spack instance you've used in the past conflicts can occur due to the `~/.spack` directory.
-The first step we recommend is deleting this directory and trying again.
+If you have multiple spack instances or have recently updated spack things can get a little mangled.
```
tasmith4: I think it’s still worth mentioning one reason is that both spack instances store things in ~/.spack

start.sh
```diff
@@ -172,3 +186,17 @@ function build-env-dive() {
...
+
+# function for cleaning spack-manager and spack directories
+function sm-clean(){
+  if [[ "$1" == "-h" || "$1" == "--help" ]]; then
+    echo "This command will clean out spack and spack-manager caches and other problematic directories"
+    return
+  fi
+  echo 'Remove user cache of configs:'
+  cmd "rm -rf ~/.spack"
+  echo 'Use spack's native clean system:
```
tasmith4: Probable typo, missing single quote (actually worse than that, the single quote in “spack’s” closes the one at the beginning of the command)

tasmith4 (commented):
I wonder if this file is getting big enough that it’s worth consolidating in some way or doing it in Python

docs/general/FAQ.md
psakievich: Sure I can add that

psakievich:
I thought about consolidating or going to python.  The main thing with these functions is we want them to execute in the active shell.  `sm-clean` could probably be python but all the rest probably not.  I'd rather leave it as is for now, but I think moving things around for better organization is on the table for the future.

jrood-nrel (approved):

tasmith4 (approved):

PR #320, opened 2022-07-21
Change default Summit compiler to GCC 10.2

jrood-nrel:
Let's just make this the default on Summit if that is what Paul is using.


psakievich (commented):
Postmortem approved

PR #322, opened 2022-07-25
Use portable identifiers for bash functions

jrood-nrel:


psakievich:
Any luck on the aliases so we don't have to break backwards compatibility? If not then we should make an announcement about this and give people a little warning before merging.

tasmith4 (approved):
Love it.  I will need a heads-up though, so that I can update halley to match this.  It would be ideal if there could be a deprecation cycle where both hyphen and underscore versions work side-by-side for a week or two.

jrood-nrel:
The aliases seem to work.

tasmith4 (approved):

psakievich (approved):

psakievich:
@tasmith4 why don't you pull the trigger on merging this when you are ready?  Also please post a notice in the spack-manager slack channel to give a heads up to people.

tasmith4:
@psakievich sure.  what deprecation timeframe do you want to announce for the hyphenated aliases?

jrood-nrel:
I say we keep the hyphen aliases.

psakievich:
If we want to keep the hyphens long term then I think we should append all the shell functions with `_` i.e. `_cd_sm` so there aren't two instances that pop up on tab completion. That should make the transition unnoticeable to the user base. Thoughts? I like the idea of keeping  them around too. I'd rather not change my command line usage of the hyphened commands.

tasmith4 (changes requested):
while testing halley I noticed a typo in the doc changes for this -- some of the docs have changed `spack manager create-env` to `spack manager create_env`.  I think this is incorrect because the actual python files didn't change, and I also think we don't want to change the hyphens to underscores in the python functions because spack uses hyphens (e.g. `spack build-env`) and I think we should be consistent.

tasmith4:
Besides the create-env thing, halley turns out to not depend on any of the hyphenated shell convenience functions, so I no longer care when this goes in if it only applies to shell functions and not python functions.

docs/user_profiles/developers/developer_workflow.md
```diff
@@ -117,9 +117,9 @@ mkdir test && cd test
 git clone --recursive --branch main git@github.com:Exawind/exawind-driver.git exawind
 git clone --recursive --branch master git@github.com:Exawind/nalu-wind.git
 git clone --recursive --branch main git@github.com:Exawind/amr-wind.git
-quick-create-dev -s exawind@main amr-wind@main nalu-wind@master
-+ spack-start
-+ spack manager create-dev-env -s exawind@master amr-wind@main nalu-wind@master
+quick_create_dev -s exawind@main amr-wind@main nalu-wind@master
++ spack_start
++ spack manager create_dev_env -s exawind@master amr-wind@main nalu-wind@master
```
tasmith4: probable typo

docs/user_profiles/developers/developer_workflow.md
```diff
@@ -72,8 +72,8 @@ you want them, or run spack install.
 The base command and it's help are echoed below:
     
 
-+ spack manager create-dev-env -h
-usage: spack manager create-dev-env [-h] [-m MACHINE] [-d DIRECTORY | -n NAME] [-y YAML] [-s SPEC [SPEC ...]]
++ spack manager create_dev_env -h
+usage: spack manager create_dev_env [-h] [-m MACHINE] [-d DIRECTORY | -n NAME] [-y YAML] [-s SPEC [SPEC ...]]
```
tasmith4: probable typos

start.sh
```diff
@@ -76,21 +76,21 @@ The base command and it's help are echoed below:
...
+    printf "\nERROR: Exiting quick_create prematurely\n"
     return 1
   fi
-  EPATH=$(cat $SPACK_MANAGER/.tmp/created_env_path.txt)
+  EPATH=$(cat ${SPACK_MANAGER}/.tmp/created_env_path.txt)
   cmd "spack env activate --dir ${EPATH} --prompt"
 }
 
-# same as quick-create but calls create-env-dev instead
+# same as quick_create but calls create_env_dev instead
```
tasmith4: probable typo carried through from original:  should be `create-dev-env` as per lines 103/107

tasmith4 (changes requested):
found a few more typos in the docs.  I think this is it

tasmith4 (approved):

psakievich:
> If we want to keep the hyphens long term then I think we should append all the shell functions with `_` i.e. `_cd_sm` so there aren't two instances that pop up on tab completion. That should make the transition unnoticeable to the user base. Thoughts? I like the idea of keeping them around too. I'd rather not change my command line usage of the hyphened commands.

I think we should do this unless you guys are actively opposed to it. This would also make it so we don't need to change the documentation at all.

jrood-nrel:
Hmm, the aliases seemed to work on my laptop, but they're not working here as I expect.

jrood-nrel:
Looks like hyphens may not be portable in aliases either.

psakievich:
Closing based on our discussions.

PR #323, opened 2022-07-28
Update spack

### LINKED ISSUE #315: Trilinos install for older commits will fail due to kokkos patch
I've noticed that if the current patches we use fail then we will be unable to do develop builds.  The example of this is trilinos with sufficiently old commits fail due to our kokkos patch.  This is an issue in spack.  We should wait until this matter get resolves and then update the spack submodule.  

I've created a PR in spack. See https://github.com/spack/spack/pull/31593

psakievich:
closes #315

### DIFF
--- a/spack-scripting/tests/conftest.py
+++ b/spack-scripting/tests/conftest.py
@@ -10,9 +10,9 @@ import os
 
 import pytest
 
-from spack.cmd import parse_specs
 import spack.paths
 import spack.repo
+from spack.cmd import parse_specs
 from spack.spec import Spec
 from spack.test.conftest import *  # noqa: F401
 
--- a/spack-scripting/tests/conftest.py
+++ b/spack-scripting/tests/conftest.py
@@ -10,6 +10,7 @@ import os
 
 import pytest
 
+from spack.cmd import parse_specs
 import spack.paths
 import spack.repo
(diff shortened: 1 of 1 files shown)


psakievich:
@tasmith4 your unit tests are in violation of a new property in spack core.  I've been looking into this but I can't find a good fix

tasmith4:
@psakievich what new property did they add, and how do the tests violate it?  Is it absolutely certain we must change the tests, or can we push back on the spack team at all?  If we absolutely have to change them, can we get help on how to do it the right way from the spack team?

psakievich:
They made it so you can't pull the package off the spec unless the spec is concrete.  This breaks your tests pretty badly though because we can't afford to concretize nalu-wind that many times, and even if we do they can't concretize on all the platforms.  For example `+cuda` is not allowed on darwin.

I can bring it up on slack with them.

psakievich:
@tasmith4 I put in the fix suggested by @alalazo.  It is a little hacky. I like the idea of adding the ability to mock the platform. Then we could do full regression test level concretizations for all platforms too.  But that will have to come in a later PR for me.

tasmith4 (approved):
yeah it does feel hacky but if that's where we're at ...

PR #327, opened 2022-08-11
add sm-clean to the nightly testing

### LINKED ISSUE #325: Updated spack has broken nightlies
After #323, all SNL nightlies are failing with the following lines in the log:

```
Concretizing environment...
+ spack concretize -f
==> Starting concretization pool with 5 processes
==> Error: errors occurred during concretization of the environment:
Package 'armpl' not found.
Package 'armpl' not found.
Package 'armpl' not found.
Package 'armpl' not found.
Package 'armpl' not found.
```

Presumably something changed with how our packages concretize in the newer spack, although it's unclear to me whether it's a spack packaging bug or a weird interaction between the concretizer and our system-specific configs.

@psakievich can you take a look?  Seems like we might want to revert #323 unless this is really easy to fix.

FYI @jrood-nrel, looks like this may also be affecting NREL nightlies.

tasmith4:
Closes #325

### DIFF
--- a/scripts/run-exawind-nightly-tests.sh
+++ b/scripts/run-exawind-nightly-tests.sh
@@ -40,6 +40,9 @@ fi
 printf "\nActivating Spack-Manager...\n"
 cmd "source ${SPACK_MANAGER}/start.sh && spack-start"
 
+printf "\nCleaning Spack installation...\n"
+cmd "sm-clean"
+
 printf "\nGenerating test script for submission...\n"
 EXAWIND_TEST_SCRIPT=${SPACK_MANAGER}/scripts/exawind-tests-script.sh
 cat > ${EXAWIND_TEST_SCRIPT} << '_EOF'
@@ -107,9 +110,6 @@ cmd "spack concretize -f"
 #else
 #  cmd "spack develop -p ${DEVELOP_SPEC_DIR} --clone hypre@develop"
 #fi
-
-printf "\nClean old stage directories...\n"
-cmd "spack clean -s || true"
 
 printf "\nTests started at: $(date)\n\n"
 printf "spack install \n"


jrood-nrel:
I don't think you want to update Spack here.

scripts/run-exawind-nightly-tests.sh
```diff
@@ -108,9 +111,6 @@ cmd "spack concretize -f"
 #  cmd "spack develop -p ${DEVELOP_SPEC_DIR} --clone hypre@develop"
 #fi
 
-printf "\nClean old stage directories...\n"
-cmd "spack clean -s || true"
```
jrood-nrel, line 112: Can the `sm-clean` stay here?

tasmith4: I'm open to suggestions on where to move it, but I think it has to come before the concretize (since the error we saw the other night, that was fixed by sm-clean, happened at concretization)

psakievich: It's probably best to have it as early as possible. Definitely before concretization.

jrood-nrel: I thought STK wanted the stage directory to persist after the tests ran.

jrood-nrel: Oh I'm all screwed up. I think this makes sense.

jrood-nrel (dismissed):
I'll approve this when the spack submodule is reverted.

tasmith4:
yeah @jrood-nrel you're absolutely right, sorry I missed that comment at first, no idea how the spack submodule change got into here.  also confused why the github autolabeler didn't tag it with "update-spack" ...

psakievich (approved):

PR #328, opened 2022-08-15
add .cache removal to sm-clean

tasmith4:
as per the discussion in #325

### DIFF
--- a/start.sh
+++ b/start.sh
@@ -195,6 +195,8 @@ function sm-clean(){
   fi
   echo "Remove user cache of configs:"
   cmd "rm -rf ~/.spack"
+  echo "Remove Spack-Manager cache:"
+  cmd "rm -rf ${SPACK_MANAGER}/.cache"
   echo "Use the native Spack clean system":
   cmd "spack clean --all"
   echo "Cleaning out pycache from spack-manager repos:"


start.sh
```diff
@@ -195,6 +195,8 @@ function sm-clean(){
   fi
   echo "Remove user cache of configs:"
   cmd "rm -rf ~/.spack"
```
jrood-nrel, line 197: I never noticed this. We should avoid removing `downloads` stored here.

start.sh
```diff
@@ -195,6 +195,8 @@ function sm-clean(){
...
+  echo "Remove Spack-Manager cache:"
+  cmd "rm -rf ${SPACK_MANAGER}/.cache"
```
jrood-nrel, line 199: Maybe it can be always assumed `${SPACK_MANAGER}` is populated here, but before using `rm -rf` in a script, you might want to check the variable exists. So:
```
 if [ ! -z "${SPACK_MANAGER}" ]; then
  cmd "cd ${SPACK_MANAGER} && rm -rf ${SPACK_MANAGER}/.cache && cd -"
fi

start.sh
tasmith4: I don't know enough about the `~/.spack` directory, although I'm also somewhat uncomfortable with the concept that `sm-clean` blows away stuff that's there when potentially a user who has multiple spack installs wants to keep some of that for other installs not related to the Spack-Manager instance that needs to be cleaned.  @jrood-nrel can you start a new issue to track this discussion?

start.sh
tasmith4: good point.  fix coming

tasmith4: Ok just looking at this a little deeper -- to source this file we already go through a check requiring `SPACK_MANAGER` to exist.  So it seems like if this function is available to be called, then we already know `SPACK_MANAGER` exists and it's kind of redundant to do another check.  @jrood-nrel is there another scenario I'm missing?

tasmith4: the check is at line 19 of this file

jrood-nrel: Ok it's fine the way it is then.

start.sh
jrood-nrel: @psakievich Maybe we should use a `~/.spack_downloads` location for the downloads instead of `~/.spack`. I'm actually fine never having `~/.spack` exist for anyone. It causes more problems than helps.

jrood-nrel: Anyway that's not part of this PR.

jrood-nrel (approved):

PR #334, opened 2022-08-18
switch back to gcc 9 for SNL nightlies to avoid configuration issue

tasmith4:


psakievich (approved):
LGTM. Don't worry about those style issues unless you're interested. They are residuals from spack changing their formatting guidelines.  I'll address it soon enough

Issue #333, opened 2022-08-18
New Syntax in Spack to apply constraints in `packages.py`


psakievich:
https://spack.readthedocs.io/en/latest/build_settings.html#package-requirements

This is a PSA for the next time we update spack. Which should be soon. 

Once https://github.com/spack/spack/pull/31558 merges we should update again

Issue #314, opened 2022-07-14
Can we reduce the number of SNL configs


psakievich:
We currently have 3 SNL configs for CEE machines: `cee` `ascicgpu` and `ascic`

The original reason for splitting out `ascicgpu` from `cee` was that the gpu version needed generic architecture to turn off SIMD when compiling with gcc and the sierra modules we were relying on had a different configuration for cuda compilers.

Both of those issues are now resolved, and I think we should evaluate combining them under the `cee` umbrella.

On a separate, but related note, we should probably move `snl-hpc` to be `tlcc2` to become more aligned with SNL lingo.  All the hpc's we use on this project fall under the `tlcc2` umbrella.

Thoughts @tasmith4 @ddement @ndevelder?

tasmith4:
I think combining cee/ascic/ascicgpu into cee makes sense, especially since it's ok to have things like cuda in the packages.yaml even if they're not used in CPU builds.  I don't think it will be completely trivial, since they're not carbon copies of each other, but I agree it's the right path forward.

I also agree with renaming snl-hpc to tlcc2, since the packages come from CDE's tlcc2 directory anyway.  However, I'll note that we'll need to consider the possibility that we want to run on CTS-1 platforms and/or vortex (ATS-2 analog) at some point.

Also, the nightly testing uses the machine name "ascic" or "ascicgpu" to do some logic (including selection of which lines to run and how to run them, and what variables to set in the ctest args list).  If the machine name needs to be the same as the config directory name then we'll need to think more deeply about how to do that differentiation a different way, although I still think it's a good move to combine the configs.

PR #336, opened 2022-08-25
update SNL configs to add buildable: false on MPI/Cuda packages

tasmith4:
This should be temporary -- related work for #333 or #314 may yield a better permanent solution for the packages.yaml callouts.


configs/ascic/packages.yaml
```diff
@@ -18,3 +18,4 @@ packages:
       prefix: /projects/wind/system-spack/opt/spack/linux-rhel7-x86_64/gcc-9.3.0/mpich-3.4.2-4h2muy6jlgfdahehxmxa4ybndfdb6gx2/
     - spec: mpich@3.4.2%gcc@8.3.1
```
psakievich, line 19:
```suggestion
    - spec: mpich@3.4.2
```
Does this still concretize correctly if we drop the compiler spec?

psakievich (approved):

PR #339, opened 2022-08-31
Hypre build variants

PaulMullowney:
When building with rocm, one needs to build hypre with rocm. Adding device blas libraries as variants.


psakievich:
@PaulMullowney I'm assuming you are adding these as variants because you still want the option to use non-device blas?  If not then we don't need to add them as variants. You can just trigger on `+cuda/+rocm`.

repos/exawind/packages/hypre/package.py
```diff
@@ -7,6 +7,8 @@ class Hypre(bHypre):
     phases = ['autoreconf', 'distclean', 'configure', 'clean', 'build', 'install']
 
     variant('gpu-aware-mpi', default=False, description='Use gpu-aware mpi')
+    variant('rocblas', default=False, description='use rocblas')
+    variant('cublas', default=False, description='use cublas')
```
psakievich, line 11:
```suggestion
    variant('cublas', default=False, description='use cublas')
    conflict('+cublas', when='~cuda', msg='cublas requires cuda to be enabled')
    conflict('+rocblas', when='~rocm', msg='rocblas requires rocm to be enabled')
```
We should add some safety rails to this.

psakievich (changes requested):

repos/exawind/packages/hypre/package.py
PaulMullowney: I like the safety checks

PaulMullowney:
> @PaulMullowney I'm assuming you are adding these as variants because you still want the option to use non-device blas? If not then we don't need to add them as variants. You can just trigger on `+cuda/+rocm`.

Actually, you don't get them when triggering +cuda/+rocm. Hypre does not do this by default. I want the option of using them.

repos/exawind/packages/hypre/package.py
PaulMullowney: I see the following in the tests:
rhodes failed to concretize. Additional exception: name 'conflict' is not defined

psakievich: whoops it's `conflicts` with an `s` my bad.

psakievich (approved):

PR #352, opened 2022-09-20
WIP: Hip porting

PaulMullowney:
This branch has some of the mods necessary for building on Crusher. It needs to be cleaned up and several outstanding issues need to be resolved.

1) It was necessary to add --hip-link to the link line of nalu via the nalu-wind CMakeLists.txt file (i.e. target_link_libraries(${nalu_ex_name} PRIVATE nalu -fgpu-rdc --hip-link) 
This is a hack that needs to be fixed through spack-manager.

2) The build only succeeds in Release mode. RelWithDebInfo and Debug builds fail at the link stage with:
unhandled SGPR spill to memory
This may be a known compiler bug that AMD is actively trying to fix. In the meantime, we should try to warn or prevent people from building with debug symbols.

### DIFF
--- a/repos/exawind/packages/nalu-wind/package.py
+++ b/repos/exawind/packages/nalu-wind/package.py
@@ -27,11 +27,13 @@ class NaluWind(bNaluWind, ROCmPackage):
             description='Enable SIMD in STK')
     variant('ninja', default=False,
             description='Enable Ninja makefile generator')
+    variant('gpu-rdc', default=False,
+            description='Enable gpu-rdc for testing')
 
     depends_on('trilinos gotype=long')
 
     for arch in ROCmPackage.amdgpu_targets:
-        depends_on('trilinos@13.4.0.2022.10.27: ~shared+exodus+tpetra+muelu+belos+ifpack2+amesos2+zoltan+stk+boost~superlu-dist~superlu+hdf5+shards~hypre+gtest+rocm~rocm_rdc amdgpu_target=%s' % arch, when='+rocm amdgpu_target=%s' % arch)
+        depends_on('trilinos@13.4.0.2022.10.27: ~shared+exodus+tpetra+muelu+belos+ifpack2+amesos2+zoltan+stk+boost~superlu-dist~superlu+hdf5+shards~hypre+gtest+rocm amdgpu_target=%s' % arch, when='+rocm amdgpu_target=%s' % arch)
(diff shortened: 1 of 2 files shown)


repos/exawind/packages/exawind/package.py
```diff
@@ -109,7 +109,7 @@ def setup_build_environment(self, env):
             env.append_flags("CXXFLAGS", "-fsanitize=address -fno-omit-frame-pointer -fsanitize-blacklist={0}".format(join_path(self.package_dir, 'sup.asan')))
         if '+rocm+amr_wind_gpu~nalu_wind_gpu' in self.spec:
             # Manually turn off device defines to solve Kokkos issues in Nalu-Wind headers
-            env.append_flags("CXXFLAGS", "-U__HIP_DEVICE_COMPILE__ -DDESUL_HIP_RDC")
+            env.append_flags("CXXFLAGS", "-D__HIP_DEVICE_COMPILE__ -DUSE_HIP_RDC")
```
tasmith4: Can we just delete this whole if statement or do we really really need it?  Specifying `-D__HIP_DEVICE_COMPILE__` manually sounds non-robust, and I think `-DUSE_HIP_RDC` should be passed through Kokkos.

@psakievich @jrood-nrel I never build amr-wind, what do you guys think?

jrood-nrel: This is necessary in the mixed device case.

tasmith4: What specifically is necessary, i.e. how should this line be modified, if at all?

jrood-nrel: You're right. This line shouldn't be modified. This case is separate from this PR which only involves `+nalu_wind_gpu`.

jrood-nrel: When amr-wind is built for AMD GPUs, the exawind-driver's cmake pulls in these defines which need to be off in Nalu-Wind so the Kokkos headers don't error.

tasmith4: got it, I'll put up a commit here to remove this change

tasmith4:
```suggestion
            env.append_flags("CXXFLAGS", "-U__HIP_DEVICE_COMPILE__ -DDESUL_HIP_RDC")
```

tasmith4:
Closing because all the main changes have merged from other PRs.  I think we can submit new targeted PRs as we identify new issues and fix them.

PR #353, opened 2022-09-20
add print statement to avoid segfault in openfast

tasmith4:


psakievich (approved):
Thanks for tracking this rather unsatisfying error/solution combo down.

tasmith4:
Yeah thanks for your help

PR #358, opened 2022-09-24
Modify Snapshots to use GitHash versions

psakievich:
Update snapshot creation process to use git hashes as versions instead of develop specs.  This should save on storage and reuse quite a bit.

This should also allow for a parallel DAG build again since hypre will have separate sources for separate installs.

It should be noted that the modules don't seem like they will work anymore with the projections.  I'm not sure what to do about this yet.  It seems like we may need to do a view module pair per root spec again.

Looking back over old conversations I think the separate build directories will solve the issues we were seeing with 1 view per root spec.  However I think we should test that out in a follow on PR after we merge this.

### DIFF
--- a/scripts/snapshot_creator.py
+++ b/scripts/snapshot_creator.py
@@ -22,8 +22,10 @@ import spack.environment as ev
 import spack.main
 import spack.util.spack_yaml as syaml
 import spack.util.executable
-from spack.version import GitVersion, Version
 import spack.cmd.install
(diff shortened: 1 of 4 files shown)


eugeneswalker (approved):

psakievich:
@eugeneswalker let's keep an eye on the docker builds.  They will likely be the first thing to test this code in the wild.

PR #359, opened 2022-09-24
add awk to repo check to reduce fragility

tasmith4:
If a user had another repo with a longer namespace, the number of spaces between the 'exawind' namespace and the start of the path would change, breaking the grep call.


psakievich (commented):
Can you add a check against `$2` to make sure it is the correct path relative to the `SPACK_MANAGER` env variable?

tasmith4:
> Can you add a check against `$2` to make sure it is the correct path relative to the `SPACK_MANAGER` env variable?

@psakievich done.

psakievich (approved):

jrood-nrel (approved):

PR #361, opened 2022-10-03
Use same diff tolerances for GPU test platforms

psakievich:


tasmith4 (approved):
@jrood-nrel @psakievich definitely go ahead with this PR, but maybe we can have a discussion about whether this is the right place to put this option ... since we expose the tolerance options as spack package options, maybe we should be specifying this in the `env-templates` files?  On the other hand, doing it in package would allow users to automatically get these tolerances when they run outside of the regular nightly testing process.  If that's seen as a positive, maybe we should consider moving this to the `exawind.nalu-wind` package instead of `nalu-wind-nightly`?

psakievich:
@tasmith4 those are good points.  I'm okay with moving it to the main nalu-wind package.  I'd like to make it as easy as possible to reproduce the nightlies in a local build.

PR #362, opened 2022-10-04
be more specific about user cache/config paths

tasmith4:
After discovering that spack has both the `SPACK_USER_CACHE_PATH` and `SPACK_USER_CONFIG_PATH`, I realized that the first two lines of `sm-clean` are really just deleting both those directories.  This PR would make that explicit by using those env variables to specify what to delete.

In order to make this work, we need to make sure `SPACK_USER_CONFIG_PATH` is actually set in `spack-start`.  Setting it to a location inside `$SPACK_MANAGER` makes it so `sm-clean` will not accidentally blow away configs/caches needed by other spack installations.

### DIFF
--- a/scripts/run-exawind-nightly-tests.sh
+++ b/scripts/run-exawind-nightly-tests.sh
@@ -41,7 +41,12 @@ printf "\nActivating Spack-Manager...\n"
 cmd "source ${SPACK_MANAGER}/start.sh && spack-start"
 
 printf "\nCleaning Spack installation...\n"
-cmd "spack clean --all"
+  echo "Remove Spack-Manager cache:"
+  cmd "rm -rf ${SPACK_MANAGER}/.cache"
+  echo "Use the native Spack clean system":
+  cmd "spack clean --all"
+  echo "Cleaning out pycache from spack-manager repos:"
+  cmd "find ${SPACK_MANAGER}/repos -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -delete"
 
 printf "\nGenerating test script for submission...\n"
 EXAWIND_TEST_SCRIPT=${SPACK_MANAGER}/scripts/exawind-tests-script.sh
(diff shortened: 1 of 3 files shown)


jrood-nrel:
I get nervous about having an `rm` on a single variable without any checks. We need to very careful about `rm -rf` in scripts.

jrood-nrel:
I think we still want `rm -rf ~/.spack` as well.

tasmith4:
@jrood-nrel that makes sense.  Removed the changes to `sm-clean`, and also removing `sm-clean` from the nightly process, so that running the blanket `rm -rf ~/.spack` doesn't mess with other spack installs.

scripts/run-exawind-nightly-tests.sh
```diff
@@ -41,7 +41,7 @@ printf "\nActivating Spack-Manager...\n"
 cmd "source ${SPACK_MANAGER}/start.sh && spack-start"
 
 printf "\nCleaning Spack installation...\n"
-cmd "sm-clean"
+cmd "spack clean --all"
```
psakievich: This is okay, but if we intend to continue a __clean__ operation as part of the nightlies then there are a few other things in `sm-clean` we may want to preserve like clearing out the python caches inside our repo: https://github.com/psakievich/spack-manager/blob/8f3377ee5e287d8a2c4aa5867541131cfb1fb9cb/start.sh#L203

This has caused issues when updating spack in the past.  The code suggestion is everything in `sm-clean` but the `rm -rf ~/.spack` command.  Overall I think it would be easier to add a flag to `sm-clean` which will cause the command to omit removing `~/.spack`.

```suggestion
  echo "Remove Spack-Manager cache:"
  cmd "rm -rf ${SPACK_MANAGER}/.cache"
  echo "Use the native Spack clean system":
  cmd "spack clean --all"
  echo "Cleaning out pycache from spack-manager repos:"
  cmd "find ${SPACK_MANAGER}/repos -type f -name '*.py[co]' -delete -o -type d -name __pycache__ -delete"
```

psakievich (approved):

tasmith4:
@psakievich my primary resistance to adding a flag to sm-clean is that I'm not very comfortable with either argument parsing or shell functions in bash.  If you have some pointers/code examples I'd be happy to implement.

psakievich:
No worries. We can do it in another PR.  I would have put it in place with the suggestions feature, but I think that only works with the files that have been edited as part of the PR.

PR #368, opened 2022-10-07
Use ninja for high level exawind packages since fortran was removed

jrood-nrel:
I haven't tested this yet.

### DIFF
--- a/repos/exawind/packages/amr-wind/package.py
+++ b/repos/exawind/packages/amr-wind/package.py
@@ -25,6 +25,9 @@ class AmrWind(bAmrWind):
     depends_on('hdf5+mpi', when='+hdf5+mpi')
     depends_on('h5z-zfp', when='+hdf5')
     depends_on('zfp', when='+hdf5')
+    depends_on("ninja", type="build")
+
+    generator = "Ninja"
 
     def setup_build_environment(self, env):
         if '+asan' in self.spec:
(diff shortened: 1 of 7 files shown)


jrood-nrel:
This seems to work for me. Hoping for a better experience this time around without fortran.

repos/exawind/packages/amr-wind/package.py
```diff
@@ -25,6 +25,9 @@ class AmrWind(bAmrWind):
     depends_on('hdf5+mpi', when='+hdf5+mpi')
     depends_on('h5z-zfp', when='+hdf5')
     depends_on('zfp', when='+hdf5')
+    depends_on("ninja", type="build")
+
+    generator = "Ninja"
```
psakievich:
```suggestion
    @property
    def generator(self):
          if '+ninja' in self.spec:
              return "Ninja"
          else:
              return "Unix Makefiles"
```
does this work to make ninja optional

repos/exawind/packages/amr-wind/package.py
```diff
@@ -25,6 +25,9 @@ class AmrWind(bAmrWind):
     depends_on('hdf5+mpi', when='+hdf5+mpi')
     depends_on('h5z-zfp', when='+hdf5')
     depends_on('zfp', when='+hdf5')
+    depends_on("ninja", type="build")
```
psakievich:
```suggestion
    depends_on("ninja", type="build", when='+ninja')
    variant('ninja', default=False)
```

repos/exawind/packages/amr-wind/package.py
jrood-nrel: Oh cool, yeah let me try that.

psakievich (approved):

PR #376, opened 2022-10-20
Nalu-Wind cxxstd

psakievich:
I'm wondering if we should just remove this variant all together.  If we add cxx17 specific features we will no longer be able to build with 14.


jrood-nrel:
We should remove it in the future and force 17.

psakievich:
Another thing we can do is keep the variant but only allow the 17 option. Then when we are looking at moving to 20 we will have the flag to switch in place.  Also hopefully by then we will have releases.

tasmith4 (approved):

ldh4:
I also suggest removing the variant and force 17 once Trilinos updates to 17 and Nalu-wind becomes stable with it.
I can't think of a case where we will be needing to use this option other than in an experimental build to try out with more updated Cpp version.

PR #381, opened 2022-10-27
Hypre does not need UVM

PaulMullowney:


repos/exawind/packages/nalu-wind/package.py
```diff
@@ -21,7 +21,6 @@ class NaluWind(bNaluWind, ROCmPackage):
     variant('ninja', default=False,
             description='Enable Ninja makefile generator')
 
-    depends_on('hypre+unified-memory', when='+hypre+cuda')
```
psakievich, line 24: Should this be 
```python
    depends_on('hypre+unified-memory', when='+hypre+cuda+uvm')
```    
or it never depends on `uvm`?

PaulMullowney: There is no uvm variant. We could add one.

psakievich: Meh. Let's not unless we need it.

psakievich (approved):

PR #382, opened 2022-10-28
promote Trilinos develop to stable

tasmith4:
@psakievich @jrood-nrel @PaulMullowney ... seems like from the slack discussion we should go ahead with this.  Let me know if you disagree or think we need to test anything else first.

### DIFF
--- a/repos/exawind/packages/trilinos/package.py
+++ b/repos/exawind/packages/trilinos/package.py
@@ -10,8 +10,8 @@ from spack.pkg.builtin.trilinos import Trilinos as bTrilinos
 import os
 
 class Trilinos(bTrilinos):
-    # develop@6-5-2022
-    version('stable', commit='7498bcb9b0392c830b83787f3fb0c17079431f06')
+    # develop@10-27-2022
+    version('stable', commit='da54d929ea62e78ba8e19c7d5aa83dc1e1f767c1')
     variant('stk_unit_tests', default=False,
             description='turn on STK unit tests')
     variant('stk_simd', default=False,


PaulMullowney:
I think we should definitely do this. We're gonna have some failing Trilinos tests on multiple devices but hopefully this will encourage people to fix. 

One thing to note, we no longer need to patch trilinos with this promotion.

psakievich:
So I'm for this, but I'm not sure I want to keep pushing `stable`.  The issue is if I install a new build and it is using `trilinos@stable` it's not clear if it is current.  I'm wondering if we should pursue a versioning strategy like Alegra, use `git.[hash]=stable` in the `packages.yaml` or something along those lines so we have clearer definitions between builds.

tasmith4:
I don't have an opinion on pushing `stable` vs a git hash versioning scheme, I'm happy to go with whatever you all like best.  @psakievich or anyone else, if you want to edit this PR to implement a versioning scheme, please feel free.

psakievich:
I'll put up a pr on a versioning scheme today. I'm thinking just the date `2022.10.24` or something. We'll need to update the nalu-wind core package too.

repos/exawind/packages/trilinos/package.py
```diff
@@ -10,8 +10,8 @@
 import os
 
 class Trilinos(bTrilinos):
-    # develop@6-5-2022
-    version('stable', commit='7498bcb9b0392c830b83787f3fb0c17079431f06')
+    # develop@10-27-2022
+    version('stable', commit='da54d929ea62e78ba8e19c7d5aa83dc1e1f767c1')
```
psakievich, line 14:
```suggestion
    version('13.4.0.2022.10.27', commit='da54d929ea62e78ba8e19c7d5aa83dc1e1f767c1')
    version('13.2.0.2022.06.05', commit='7498bcb9b0392c830b83787f3fb0c17079431f06')
```

psakievich: I'd like to move to actual versions so we maintain a set of commit histories in the package. I realize this may break things. Debate is welcome. Depends on https://github.com/spack/spack/pull/33620

psakievich (changes requested):

repos/exawind/packages/trilinos/package.py
psakievich: I'll open a separate PR for this

tasmith4:
closing since #387 supersedes and is merged

PR #384, opened 2022-10-28
disable Fortran in hypre

tasmith4:
Part 2 of breakup of #352.  I think this is not strictly required for ROCm, but since it was on @PaulMullowney's branch I presume it is valuable and didn't want it to get lost.

@psakievich @jrood-nrel


PaulMullowney (commented):
I removed these in my most recent build. We don't need these changes.

tasmith4:
> I removed these in my most recent build. We don't need these changes.

Ok great, I will close this then.

PR #385, opened 2022-10-28
changes to nalu-wind package to get building on Crusher

tasmith4:
Part 3 of #352.  Just a note here so it doesn't get lost, #352 included a change to some AMR-Wind-specific flags in the exawind package.  As per my discussion with @jrood-nrel on that issue, I'm not including that change in any of these pieces of #352.

@PaulMullowney @psakievich @jrood-nrel @ddement @djglaze


PaulMullowney (commented):
We don't seem to need
depends_on('hypre +rocm ...' ... )
I'm not sure why this is the case.

repos/exawind/packages/nalu-wind/package.py
```diff
@@ -21,11 +21,13 @@ class NaluWind(bNaluWind, ROCmPackage):
     variant('ninja', default=False,
             description='Enable Ninja makefile generator')
 
+    for _arch in ROCmPackage.amdgpu_targets:
+        depends_on('hypre+rocm amdgpu_target={0}'.format(_arch), when='+hypre+rocm amdgpu_target={0}'.format(_arch))
```
tasmith4:
```suggestion
```

so @PaulMullowney we can just delete these lines?

PaulMullowney: I think so. I would like input from @jrood-nrel

PaulMullowney: I'm rebuilding right now with I think is the correct setup. If successful (i.e. the hip gets all the way to the nalu link phase), I'll push to my branch and we can reevaluate this.

jrood-nrel: Unless I'm missing something, I don't know how else rocm and its target will get propagated to hypre without this.

repos/exawind/packages/nalu-wind/package.py
```diff
@@ -70,6 +74,8 @@ def cmake_args(self):
         if '+rocm' in self.spec:
             cmake_options.append('-DCMAKE_CXX_COMPILER={0}'.format(self.spec['hip'].hipcc))
             cmake_options.append(define('ENABLE_ROCM', True))
+            targets = spec.variants['amdgpu_target'].value
+            cmake_options.append('-DGPU_TARGETS=' + ';'.join(str(x) for x in targets))
```
jrood-nrel: In general I think we should start using the builtin define methods more instead of doing them explicitly. We use them sometimes, but you can see how I tried to always use them in openfast https://github.com/spack/spack/blob/develop/var/spack/repos/builtin/packages/openfast/package.py

tasmith4: makes sense, will change

tasmith4: @jrood-nrel this should be fixed now

tasmith4:
@psakievich @PaulMullowney @jrood-nrel I think this is ready now, any objections to merging?

repos/exawind/packages/nalu-wind/package.py
```diff
@@ -68,17 +68,17 @@ def cmake_args(self):
...
+            cmake_options.append(self.define('ENABLE_TESTS', True))
 
         if '+rocm' in self.spec:
-            cmake_options.append('-DCMAKE_CXX_COMPILER={0}'.format(self.spec['hip'].hipcc))
-            cmake_options.append(define('ENABLE_ROCM', True))
+            cmake_options.append(self.define('CMAKE_CXX_COMPILER', self.spec['hip'].hipcc))
+            cmake_options.append(self.define('ENABLE_ROCM', True))
             targets = spec.variants['amdgpu_target'].value
-            cmake_options.append('-DGPU_TARGETS=' + ';'.join(str(x) for x in targets))
+            cmake_options.append(self.define('GPU_TARGETS', ';'.join(str(x) for x in targets)))
```
PaulMullowney: Where is the builtin define method?

jrood-nrel (approved):
I'm fine merging this. Feels like there is a disconnect for rdc by using `CXXFLAGS` instead of some Kokkos option or the trilinos rdc variant though.

repos/exawind/packages/nalu-wind/package.py
psakievich: It's defined in the parent class for cmake packages in spack.

tasmith4:
> I'm fine merging this. Feels like there is a disconnect for rdc by using `CXXFLAGS` instead of some Kokkos option or the trilinos rdc variant though.

Agreed.  I think that is one of the things we should clean up once the HIP software stack has stabilized.

PR #387, opened 2022-10-31
Add numeric versions to Trilinos and remove stable version

psakievich:
If a user already has `trilinos@stable` in it the `--reuse` will pick it up, but using `--fresh` will pick up the new "stable" version of trilinos

### DIFF
--- a/configs/base/packages.yaml
+++ b/configs/base/packages.yaml
@@ -26,7 +26,7 @@ packages:
   yaml-cpp:
     version: [0.6.3]
   trilinos:
-    version: [stable, develop]
+    version: [13.4.0.2022.10.27, develop]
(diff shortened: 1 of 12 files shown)


psakievich:
supersedes portions of #382

tasmith4 (approved):
I think this is reasonable.  But I don't have strong opinions on how the stable versions of Trilinos are named so would like to hear from @jrood-nrel also.

jrood-nrel:
I'm fine with the naming updates. I just realized the proposed Trilinos commit doesn't build with the Intel compiler so I think we should wait on upgrading Trilinos until that is fixed.

psakievich:
closes #382

psakievich:
@jrood-nrel I can't conditionally set the trilinos config on eagle like I thought.  Spack still needs to implement this functionality.  However, maintaining the versions and cxx14 option gives an intel path forward for now.  I've updated the rhodes testing to show how to do this.

jrood-nrel (approved):
Is the Spack submodule update intentional?

psakievich:
@jrood-nrel yes it is to allow for https://github.com/spack/spack/commit/6990cf76a0d167db844373d8b643d42cd3282049

Which enables us to use this trilinos versioning scheme.

PR #397, opened 2022-11-02
Refactor snapshot code to become a spack-command

psakievich:
Major refactor of snapshots.  Now it will be a command.  I've switched it to use an environment template that we will modify in the command to add things like the date. 

Some major changes
- `spack manager snapshot` is now a command
- remove the option to use develop specs (we don't need this with git refs)
- install tree for the snapshot is now placed inside the environment directory
   **Pro:** installation is separate from the spack database and so we don't need to worry about view types anymore.  The binaries will be safe even if we blow away the spack installs.  It will only be deleted if we remove the environment directory where the binaries are installed
  **Con:** installations will end up installing the ENTIRE DAG every time unless the `includes.yaml` has externals/upstreams.  This is the case on all of our production machines so I think it is a small con.
- Module will now use lmod and use a hierarchy to filter the available modules
  This is not as ideal as a single module for each root, but it makes the command more flexible and requires less work on our end since the LMOD infrastructure is supported in spack


psakievich:
I think this is mostly there. @eugeneswalker @jrood-nrel if you have any input on the modules stuff that would be helpful. I need to test those out and make sure they look good on the HPC's.  That is really all that is left in terms of nuts and bolts.  

It would be nice to have more tests but they are expensive and I don't have time to figure out how to mock/monkeypatch all the things that need to be mocked right now.

psakievich:
Working on refactoring some of these features into separate spack commands #430 #429

psakievich:
#432 combines this and #430.

PR #401, opened 2022-11-04
Patch STK CouplingVersions.cpp to allow building to succeed on Eagle

jrood-nrel:
Probably a better fix for this, but I'm hoping to at least do this temporarily until a better fix is made.

### DIFF
--- a/repos/exawind/packages/trilinos/package.py
+++ b/repos/exawind/packages/trilinos/package.py
@@ -22,6 +22,7 @@ class Trilinos(bTrilinos):
             description="Turn on address sanitizer")
 
     patch("kokkos_zero_length_team.patch")
+    patch("stk-coupling-versions-func-overload.patch")
 
     depends_on("ninja", type="build", when="+ninja")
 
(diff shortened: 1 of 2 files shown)


repos/exawind/packages/trilinos/package.py
```diff
@@ -22,6 +22,7 @@ class Trilinos(bTrilinos):
             description="Turn on address sanitizer")
 
     patch("kokkos_zero_length_team.patch")
+    patch("stk-coupling-versions-func-overload.patch")
```
psakievich: Make this conditional on the spack manager machine being eagle?
@tasmith4 any idea why this overload is here? Do we need it anywhere else?

psakievich (approved):
I'll conditionally approve so you can merge after the fix in my comment or if need be in a follow on pr. I'd like to isolate the patch to eagle until we know the reason for that code.

repos/exawind/packages/trilinos/package.py
tasmith4: @psakievich @jrood-nrel I am looking into this now.  I suspect we can either get rid of the overload or refactor the public API here.  I think merging this patch only for eagle until that fix makes it into Trilinos is the right move.

tasmith4 (approved):

psakievich (approved):

PR #404, opened 2022-11-04
changes to get Trilinos building on crusher

tasmith4:
Should be the last of the changes from #352.


repos/exawind/packages/trilinos/package.py
```diff
@@ -23,7 +23,6 @@ class Trilinos(bTrilinos):
     variant("asan", default=False,
             description="Turn on address sanitizer")
 
-    patch("kokkos_zero_length_team.patch")
```
psakievich, line 26: We should keep this patch and do:
```python
patch("kokkos_zero_length_team.patch", when'@:13.3.0')
```
to support the old version of Trilinos until we remove it.

psakievich (changes requested):

repos/exawind/packages/trilinos/package.py
```diff
@@ -23,7 +23,6 @@ class Trilinos(bTrilinos):
     variant("asan", default=False,
             description="Turn on address sanitizer")
```
tasmith4, line 25:
```suggestion

patch("kokkos_zero_length_team.patch", when'@:13.3.0')

```

repos/exawind/packages/trilinos/package.py
```diff
@@ -23,7 +23,8 @@ class Trilinos(bTrilinos):
...
 
-    patch("kokkos_zero_length_team.patch")
+patch("kokkos_zero_length_team.patch", when'@:13.3.0')
```
tasmith4:
```suggestion
    patch("kokkos_zero_length_team.patch", when'@:13.3.0')
```

repos/exawind/packages/trilinos/package.py
tasmith4: ok how about that?

psakievich (approved):

repos/exawind/packages/trilinos/package.py
```diff
@@ -23,7 +23,8 @@ class Trilinos(bTrilinos):
     variant("asan", default=False,
             description="Turn on address sanitizer")
 
-    patch("kokkos_zero_length_team.patch")
+    patch("kokkos_zero_length_team.patch", when'@:13.3.0')
```
tasmith4:
```suggestion
    patch("kokkos_zero_length_team.patch", when="@:13.3.0")
```

psakievich (approved):

PR #406, opened 2022-11-04
Try out depfiles for nightly testing

psakievich:
@jrood-nrel @tasmith4 something we could test out one night in the future. This is an embarrassingly parallel build process I tried to day on a regular environment. 

https://spack.readthedocs.io/en/latest/environments.html#generating-depfiles-from-environments

### DIFF
--- a/scripts/run-exawind-nightly-tests.sh
+++ b/scripts/run-exawind-nightly-tests.sh
@@ -132,10 +132,13 @@ cmd "spack concretize -f --fresh"
 
 printf "\nTests started at: $(date)\n\n"
 printf "spack install \n"
+spack env depfile -o Makefile
 if [ "${SPACK_MANAGER_MACHINE}" == 'ascicgpu' ]; then
-  time (spack install --keep-stage)
+  # time (spack install --keep-stage)
+  time (make -j32)
 else
-  time (for i in {1..4}; do spack install --keep-stage --fresh & done; wait)
+  # time (for i in {1..4}; do spack install --keep-stage --fresh & done; wait)
+  time (make -j32)
 fi
 printf "\nTests ended at: $(date)\n"
 
--- a/scripts/run-exawind-nightly-tests.sh
+++ b/scripts/run-exawind-nightly-tests.sh
@@ -135,7 +135,7 @@ printf "spack install \n"
 spack env depfile -o Makefile
 if [ "${SPACK_MANAGER_MACHINE}" == 'ascicgpu' ]; then
   # time (spack install --keep-stage)
-  time (make -j32)
+  time (make SPACK_INSTALL_FLAGS=--keep-stage -j32)
 else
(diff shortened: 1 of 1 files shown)


tasmith4:
@psakievich looks like it is missing the actual depfile?

psakievich:
It isn't. Makefile is the depfile. It's pretty cool. When you call `make` it DAG and package level parallelisms are combined.

tasmith4:
ok yeah I don't see a problem trying this sometime.  whenever it works for @jrood-nrel is fine with me, as far as I'm concerned if it breaks our nightlies too bad we can always revert

scripts/run-exawind-nightly-tests.sh
```diff
@@ -132,10 +132,13 @@ cmd "spack concretize -f --fresh"
 
 printf "\nTests started at: $(date)\n\n"
 printf "spack install \n"
+spack env depfile -o Makefile
 if [ "${SPACK_MANAGER_MACHINE}" == 'ascicgpu' ]; then
-  time (spack install --keep-stage)
+  # time (spack install --keep-stage)
+  time (make -j32)
```
psakievich:
```suggestion
  time (make SPACK_INSTALL_FLAGS=--keep-stage -j32)
```

scripts/run-exawind-nightly-tests.sh
```diff
@@ -132,10 +132,13 @@ cmd "spack concretize -f --fresh"
...
 else
-  time (for i in {1..4}; do spack install --keep-stage --fresh & done; wait)
+  # time (for i in {1..4}; do spack install --keep-stage --fresh & done; wait)
+  time (make -j32)
```
psakievich:
```suggestion
  time (make SPACK_INSTALL_FLAGS='--keep-stage --fresh' -j32)
```

tasmith4 (approved):

jrood-nrel:
I'm fine with this if it's a `make -j4`.

psakievich:
> I'm fine with this if it's a `make -j4`.

Hmm good point. We should probably test this out on the matrix to see how it does before merging.  Let's just leave it open for a bit until we have time to test it out

psakievich:
> I'm fine with this if it's a `make -j4`.

@jrood-nrel is this out of concern that we'd be running too many concurrent tests on the machines? 4 cores seems really small.

jrood-nrel:
Sometimes Nalu is running 16 processes (build_jobs=16 on rhodes) at a time and they look like they can be 2GB per process. If we have more than 4x of those concurrently, I think it could be a problem for the test machine that has 128GB of RAM.

psakievich:
It seems like we should split the dependency build from the tests before doing this. We can unleash the full build on the dependencies and then do a linear spack install for the tests so they run 1 at a time?

jrood-nrel:
I would still like to run the test categories in parallel. The clang tests take 12 hours.

PR #409, opened 2022-11-07
Allow consistent name for SNL dashboard

psakievich:
Make it so the version is always "stable" or "develop" for the SNL dashboard filtering operation

### DIFF
--- a/repos/exawind/packages/nalu-wind-nightly/package.py
+++ b/repos/exawind/packages/nalu-wind-nightly/package.py
@@ -41,7 +41,8 @@ class NaluWindNightly(bNaluWind, CudaPackage):
         return compiler + cuda
 
     def dashboard_trilinos(self):
-        trilinos = self.name_and_version("trilinos")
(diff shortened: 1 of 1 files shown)


repos/exawind/packages/nalu-wind-nightly/package.py
```diff
@@ -41,7 +41,8 @@ def dashboard_compilers(self):
         return compiler + cuda
 
     def dashboard_trilinos(self):
-        trilinos = self.name_and_version("trilinos")
+        vstring = bNaluWind.trilinos_version_filter(self.spec["trilinos"].version)
```
tasmith4:
```suggestion
        vstring = self.trilinos_version_filter(self.spec["trilinos"].version)
```

why not this?

psakievich: It's not a member function.

tasmith4 (approved):

PR #413, opened 2022-11-08
Consolidate and rename SNL configs

psakievich:
CEE ascicgpu and ascic are all using the same configs so it seems silly to keep them separate.  
I believe this won't affect the nightly testing since the logic of if it is a gpu test or not is embedded in the `nalu_wind_snl_[cpu/gpu]_nightly_test.sh` files, and these should be launched via cron job on the specific machines.

Also rename `snl-hpc` to `tlcc2` to provide a clearer SNL description of the platforms supported.


psakievich:
Closes #314

repos/exawind/packages/nalu-wind-nightly/package.py
```diff
@@ -89,7 +89,7 @@ def ctest_args(self):
         ctest_options.extend([self.define("TESTING_ROOT_DIR", self.stage.path),
             self.define("NALU_DIR", self.stage.source_path),
             self.define("BUILD_DIR", self.build_directory)])
-        if full_machine == "eagle.hpc.nrel.gov" or machine == "ascicgpu":
```
tasmith4, line 92: not excited about either of these two similar changes.  this feels like something that is machine-dependent, not cuda-dependent.  the other one is more borderline to me (perhaps we always want tolerances to depend on `+cuda` on every cuda machine?), but I still don't love it.

spack-scripting/scripting/cmd/manager_cmds/find_machine.py
```diff
@@ -17,14 +17,14 @@ def __init__(self, test, full_machine_name):
...
-    known_hosts = ("cee", "ews", "ecw", "ecs", "hpws")
+    known_hosts = ("cee", "ews", "ecw", "ecs", "hpws", "ascic")
     for k in known_hosts:
         if k in hostname:
             return True
     return False
 
 
-def is_snl_hpc(hostname):
+def is_tlcc2(hostname):
```
tasmith4: note that ghost and attaway are CTS-1, not TLCC2.  because I think it's misleading to lump things under tlcc2 if they are not TLCC2, I prefer to either (a) stick with a catch-all snl_hpc config, (b) remove all machines that are not TLCC2 from the known_hosts list, or (c) create an is_cts1 and corresponding set of configs.

tasmith4 (commented):
In general I like this direction.  But I think it's important to realize there are two separate concepts:  which machine am I running on?, and which set of configs do I want to use?

Historically we've conflated those two by having each machine have its own config.  Now that we're moving away from that strategy, I think selecting config using "SPACK_MANAGER_MACHINE" isn't the right strategy.  Seems like instead of just having find_machine, we should have find_machine and find_config_set (or similar).  That way we can make choices most of the time based on common config set, but separate things based on machine when we need to.

repos/exawind/packages/nalu-wind-nightly/package.py
psakievich: Well this is actually the case. The only tests that run with higher tolerances are cuda tests.  All the cpu ones can run with tighter tolerances. We only run cuda tests on `ascicgpu` and `eagle`. If we were to run cpu tests there then I think we'd want those runs to have tighter tolerances on those tests.

spack-scripting/scripting/cmd/manager_cmds/find_machine.py
psakievich: ah good point. I didn't look that closely. I'm not really a huge fan of the SNL naming scheme. I'll think about how to fix this.  Reverting back to `snl_hpc` might be the way to go.

psakievich:
@tasmith4 I don't know if I fundamentally agree with you on the partitioning of machines and configs.  I would prefer to have them unified for a given project so you only have one set of configs per machine.  Historically the reason why I split `ascicgpu` off from `cee` was because we were using the sierra modules and there was a conflict in trying to support cuda and non-cuda gcc compilers in the same config. 

Spack and this project have both evolved a long way since then and since we've been running with the exact same configs for `ascicgpu` `ascic` and `cee` for several months I think we should just merge them and do our differentiation via specs. There is the question of if they should be separate because they are actually using different machines, but the `cee` and `ascic` machines are such a mixed bag that we've just generalized to the most agnostic architecture possible.  If `ascic` machines compile differently for `x86_64` architecture than the `cee` machines then I think we have a serious problem that needs to be elevated beyond our group.   I don't want to dive into specializing this unless we absolutely have to.

The one caveat I can think of here is gold files from nightly test.  That was an issue in the past when you were triggering things off the machine, but  I don't see that as an issue since you've already partitioned the golds based on which machine the tests are launched from, and that is actually independent of `spack-manager`. This could grow to be an issue in the future, but for now I think it is okay.

tasmith4 (approved):

PR #432, opened 2022-11-14
Create a snapshot spack manager command

psakievich:
combine #429 #430 #431 for use while waiting on reviews

PR #434, opened 2022-11-17
Update module syntax for eagle

psakievich:
This module file is breaking with updated commits of spack. Putting up a draft PR as a PSA.  Implementation is in #432 as well since spack is already updated there. @jrood-nrel


psakievich:
Converting it to a full PR since it passes all the tests.  Might as well merge sooner than later.

jrood-nrel (approved):

PR #437, opened 2022-11-18
Admin setup compilers and core libraries

psakievich:
Create a way to bootstrap the same compilers, mpi etc on a given system. Right now targeting `cee` and `snl-hpc`.


scripts/bootstrap_tpls/pcomps/spack.yaml
```diff
@@ -0,0 +1,27 @@
...
+  specs:
+  - gcc@10.3.0
+  - llvm@13+clang targets=x86 ~gold ^ninja@kitware
+  view: false
+  concretizer:
+    unify: false
+  packages:
+    all:
+      target: [x86_64]
+      variants: [build_type=Release]
```
jrood-nrel: I think you will want to use `requires` in these so you don't get unexpected behavior.

scripts/bootstrap_tpls/Makefile
```diff
@@ -0,0 +1,79 @@
...
+export SPACK_USER_CACHE_PATH=$(CURDIR)/.spack
+
+.SUFFIXES:
+
+.PHONY: all clean
+
+all: tpls/generated/env
+
+clean:
+	rm -rf icomp/spack.mk icomp/spack.lock icomp/compilers.yaml icomp/config.yaml icomp/packages.yaml icomp/generated icomp/.spack-env
```
jrood-nrel: I think the `-r` can be removed here for safety.

jrood-nrel: Maybe there are directories in there I guess.

psakievich: Yeah `.spack-env` is a directory

scripts/bootstrap_tpls/pcomps/spack.yaml
psakievich: For this case it _shouldn't_ matter. Each install tree is blank so there is nothing for the concretizer to choose over the suggestions we give it. But yeah `requires` is probably safer.

eugeneswalker (approved):

jrood-nrel (approved):

PR #444, opened 2022-11-23
Add documentation about using the parallel DAG

jrood-nrel:
Didn't look like we have this anywhere in the docs yet.


psakievich (changes requested):
Thanks for doing this!  Since you're already in here can you add a blurb and link about depfiles. Depfiles are getting used more and more in spack. Pretty sure they are used in Spack CI and there is a PR open to refactor bootstrapping to use environments and depfiles as well.  I think this will become the primary use case in the future.  Here is the link in the spack documentation: https://spack.readthedocs.io/en/latest/environments.html#generating-depfiles-from-environments

jrood-nrel:
Sure, let me add that.

psakievich (approved):

PR #445, opened 2022-11-25
Update cee configs

psakievich:
Use compilers and packages built from #437. We should test this with other users besides me before merging.


tasmith4 (approved):
I successfully built with this PR for both gcc@10.3.0 and clang@13.0.1

PR #449, opened 2022-11-29
Add files for building system modules at NREL

jrood-nrel:
Probably need to figure out an organization, but making this PR to make my work visible.


psakievich:
I think this is fine if you want to merge it for documentation purposes. We should figure out a consolidation process at some point. If you do want to merge then: 1) is the spack submodule change intentional, 2) can we remove/reuse redundant files like the mixed toolchain compiler tool?

psakievich (approved):

PR #454, opened 2022-12-02
Remove redundant call to set bootstrap store dir

psakievich:
We are currently calling the command to set the bootstrap store too much It leads to an unnecessary line every time the users run `spack-start`

### DIFF
--- a/scripts/spack_start.sh
+++ b/scripts/spack_start.sh
@@ -20,7 +20,7 @@ if ! $(type '_spack_start_called' 2>/dev/null | grep -q 'function'); then
   #spack clean -m
   
   # move the bootstrap store outside the user config path
-  if [[ -z $(spack config --scope site blame bootstrap | grep "root: ${SPACK_MANAGER}/.bootstrap") ]]; then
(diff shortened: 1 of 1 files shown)


scripts/spack_start.sh
```diff
@@ -20,7 +20,9 @@ if ! $(type '_spack_start_called' 2>/dev/null | grep -q 'function'); then
   #spack clean -m
   
   # move the bootstrap store outside the user config path
-  spack bootstrap root ${SPACK_MANAGER}/.bootstrap
+  if [[ -z $(spack config --scope site blame bootstrap | grep "root: ${SPACK_MANAGER}/.bootstrap") ]]; then
```
tasmith4: is `--scope site` really correct here?  since the bootstrap root is not added at site scope it seems like this might be slightly inconsistent?

psakievich:
```suggestion
  if [[ -z $(spack config blame bootstrap | grep "root: ${SPACK_MANAGER}/.bootstrap") ]]; then
```

tasmith4 (approved):

PR #459, opened 2022-12-12
Update crusher configuration

jrood-nrel:
Also switch to CCE. After using Frontier where the AMD compiler doesn't work like Crusher, I think it makes sense to switch to CCE so we will have more support and hopefully better luck transitioning to Frontier.

### DIFF
--- a/configs/base/packages.yaml
+++ b/configs/base/packages.yaml
@@ -4,31 +4,31 @@ packages:
   conduit:
     variants: ~fortran~hdf5_compat
   boost:
-    version: [1.76.0]
+    require: "@1.76.0"
     variants: cxxstd=17
   hdf5:
-    version: [1.10.7]
+    require: "@1.12.1"
     variants: +cxx+hl
   netcdf-c:
-    version: [4.7.4]
+    require: "@4.7.4"
     variants: +parallel-netcdf maxdims=65536 maxvars=524288
   openfast:
-    version: [master]
+    require: "@master"
     variants: +cxx
   parallel-netcdf:
-    version: [1.12.2]
+    require: "@1.12.2"
     variants: ~fortran
   perl:
     require: "@5.34.1"
   python:
     require: "@3.9.15"
   tioga:
-    version: [develop]
+    require: "@develop"
   hypre:
-    version: [develop]
+    require: "@develop"
     variants: ~fortran
   yaml-cpp:
-    version: [0.6.3]
+    require: "@0.6.3"
   trilinos:
     version: [13.4.0.2022.10.27, develop]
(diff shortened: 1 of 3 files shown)


PaulMullowney (commented):
Does Nalu build with this?

tasmith4:
@djglaze @ddement what do you think?

psakievich:
`Nalu-Wind+rocm` does not build with this. @ddement @djglaze and I were trying earlier today.  Another new linker issue I'm trying to understand.

PaulMullowney:
> `Nalu-Wind+rocm` does not build with this. @ddement @djglaze and I were trying earlier today. Another new linker issue I'm trying to understand.

What's the linker issue?

psakievich:
> What's the linker issue? 

`llvm-ar` is getting called with linker arguments.  This seems to be a constraint only coming from `+rocm` and the resulting `llvm` associated compiler.  It is unclear the source though. It seems to be something at the CMake level, but we combed through every module file and I can't find any setting or env variable that would cause this to behave differently than the current config using the `PrgEnv-amd` environment.  
```
   56 /opt/rocm-5.2.0/llvm/bin/llvm-ar qc libzoltan.a -fuse-ld=gold -Wl,--as-needed CMakeFiles/zoltan.dir/all/all_allo.c.o CMakeFiles/zoltan.dir/coloring/col
   55 /opt/rocm-5.2.0/llvm/bin/llvm-ar: error: unknown option f
```

PaulMullowney:
> > What's the linker issue?
> 
> `llvm-ar` is getting called with linker arguments. This seems to be a constraint only coming from `+rocm` and the resulting `llvm` associated compiler. It is unclear the source though. It seems to be something at the CMake level, but we combed through every module file and I can't find any setting or env variable that would cause this to behave differently than the current config using the `PrgEnv-amd` environment.
> 
> ```
>    56 /opt/rocm-5.2.0/llvm/bin/llvm-ar qc libzoltan.a -fuse-ld=gold -Wl,--as-needed CMakeFiles/zoltan.dir/all/all_allo.c.o CMakeFiles/zoltan.dir/coloring/col
>    55 /opt/rocm-5.2.0/llvm/bin/llvm-ar: error: unknown option f
> ```

I haven't seen that one before. Fun :(

psakievich:
@PaulMullowney can you ask around at AMD and see what kind of linker constraints there are when turning on RDC?  @jrood-nrel was able to build `amr-wind+rocm` with this config and I'm trying to figure out where the constraints could be coming from.  I'm going to see if I can reproduce the `amr-wind+rocm` build from my environment as well.

Update:
So `amr-wind` is using `llvm-ar` as well.  There are just no additional arguments being passed to it, but it is being built with shared libs.  Gonna try with static libs and see if it fails or not.

PaulMullowney:
I've poked 1 colleague. I'll let you know if he has any insight. 

I don't understand the motivation of this PR. We have, I thought a working toolchain for amr-wind on crusher. Nalu-Wind compiles and has shown some success in the past.

jrood-nrel:
I haven't been able to compile anything with `PrgEnv-amd` on Frontier like I can on Crusher, where `PrgEnv-cray` seems to act the same and succeed on both. I think we will have better luck and support with Cray compilers on Frontier.

jrood-nrel:
Turns out CCE can't successfully run stk_balance on Crusher's CPUs so I think that nullifies this PR.

PR #469, opened 2022-12-21
Apply seacas patch only to trilinos@develop

jrood-nrel:


psakievich (changes requested):
`trilinos@13.4.0.2022.10.27+rocm` will still fail to build. So the default build on crusher will still fail. Can we make the default trilinos version on the crusher config to use `@develop` as well?

psakievich (approved):

PR #477, opened 2023-01-10
update version range on STK patch

tasmith4:
@jrood-nrel we have a fix for this in Trilinos/develop now.

@psakievich Trilinos is currently at 13.5, and the fix went in late in the day on Dec. 15 (so versions Dec. 16 and later should be good).  Can you confirm that my `when=` syntax is correct to continue to apply the patch for checkouts through Dec. 15?


jrood-nrel (approved):

psakievich (approved):
LGTM

PR #482, opened 2023-01-26
Frontier

frobnitzem:
FYI: Here are some suggestions for configuring the frontier environment.  I haven't been able to test these yet though.


PaulMullowney (approved):

alanw0 (approved):
Glad to see this frontier config getting put in.
We need to establish whether nalu-wind can work with rocm 5.2...
We could also try a rocm 5.2 build on crusher to see whether our recent fixes have made that ok or not.

jrood-nrel (approved):

PR #484, opened 2023-02-01
Make trilinos-solvers optional.

alanw0:
Removed explicit activation of belos, muelu, amesos2, ifpack2 Added infrastructure for enabling/disabling them.
Also make ~trilinos-solvers default on crusher.

### DIFF
--- a/configs/crusher/packages.yaml
+++ b/configs/crusher/packages.yaml
@@ -97,6 +97,8 @@ packages:
           - openblas/0.3.17
   trilinos:
     require: "@develop"
+  nalu-wind:
+    variants: ~trilinos-solvers
   all:
     compiler:
       - clang@15.0.0
--- a/configs/base/packages.yaml
+++ b/configs/base/packages.yaml
@@ -31,7 +31,7 @@ packages:
     version: [0.6.3]
   trilinos:
     version: [13.4.0.2022.10.27, develop]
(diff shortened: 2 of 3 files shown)


configs/crusher/packages.yaml
```diff
@@ -97,6 +97,8 @@ packages:
           - openblas/0.3.17
   trilinos:
     require: "@develop"
+  nalu-wind:
+    variants: ~trilinos-solvers
   all:
```
PaulMullowney, line 102: Does this mean it's off by default? And so matrix free will also get turned off in the nalu build?

psakievich: Yes but just for crusher

psakievich (approved):

configs/crusher/packages.yaml
alanw0: Yes but here it only applies to crusher. What do you think? I was thinking we usually don't want trilinos-solvers built on crusher, but I can easily drop this if you think it should remain on by default.

PaulMullowney: Fine by me

PaulMullowney (approved):

configs/crusher/packages.yaml
alanw0: Now that I think about this, if we are going to turn trilinos-solvers off by default, we should also turn hypre on by default. Is it on or off by default?

tasmith4: pretty sure it is now on by default after our latest spack change

alanw0: Yes it appears so... Did we do that deliberately? Maybe we need an opinion from @PaulMullowney on whether hypre should be on by default. It does seem like hypre is becoming the defacto default for nalu-wind...

psakievich: This is right. We want the default every where but AMD platforms to be `+trilinos-solvers+hypre` and on AMD platforms it should be `~trilinos-solvers+hypre`.

PR #486, opened 2023-02-02
Disable Nalu-Wind unit tests on Eagle when using the Intel compiler

jrood-nrel:
I tried to get this logic into the Eagle `packages.yaml`, but the `require` doesn't have a `when` clause to only use the `unit-tests` variant when `%intel` is in use, so it has to go in `nalu-wind/package.py`.

### DIFF
--- a/repos/exawind/packages/nalu-wind/package.py
+++ b/repos/exawind/packages/nalu-wind/package.py
@@ -52,6 +52,7 @@ class NaluWind(bNaluWind, ROCmPackage):
     cxxstd=["14", "17"]
     variant("cxxstd", default="17", values=cxxstd,  multi=False)
     variant("tests", default=True, description="Activate regression tests")
+    variant("unit-tests", default=True, description="Activate unit tests")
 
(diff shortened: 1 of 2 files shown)


repos/exawind/packages/nalu-wind/package.py
```diff
@@ -87,6 +88,9 @@ def cmake_args(self):
         cmake_options.append(self.define_from_variant("CMAKE_CXX_STANDARD", "cxxstd"))
         cmake_options.append(self.define_from_variant("BUILD_SHARED_LIBS", "shared"))
 
+        if find_machine(verbose=False) == "eagle" and "%intel" in spec:
```
psakievich, line 91: This is fine, but it might be more appropriate in the eagle configs.

psakievich (approved):

repos/exawind/packages/nalu-wind/package.py
jrood-nrel: I wrote a comment about this above.

PR #489, opened 2023-02-06
Umpire hooks

PaulMullowney:
Adding hooks to exawind and amr-wind when building hypre with umpire.

### DIFF
--- a/repos/exawind/packages/amr-wind/package.py
+++ b/repos/exawind/packages/amr-wind/package.py
@@ -22,12 +22,15 @@ class AmrWind(bAmrWind):
             description="Enable HDF5 plots with ZFP compression")
     variant("ninja", default=False,
             description="Enable Ninja makefile generator")
+    variant("umpire", default=False,
+            description="Enable Umpire")
 
     depends_on("hdf5~mpi", when="+hdf5~mpi")
     depends_on("hdf5+mpi", when="+hdf5+mpi")
     depends_on("h5z-zfp", when="+hdf5")
     depends_on("zfp", when="+hdf5")
     depends_on("ninja", type="build", when="+ninja")
+    depends_on("hypre+umpire", when="+umpire")
 
     @property
     def generator(self):
@@ -55,6 +58,10 @@ class AmrWind(bAmrWind):
 
         if spec.satisfies("dev_path=*"):
             cmake_options.append(self.define("CMAKE_EXPORT_COMPILE_COMMANDS", True))
+
+        if "+umpire" in self.spec:
+            cmake_options.append(self.define_from_variant("AMR_WIND_ENABLE_UMPIRE", "umpire"))
+            cmake_options.append(self.define("UMPIRE_DIR", self.spec["umpire"].prefix))
 
         if "+cuda" in self.spec:
             targets = self.spec.variants["cuda_arch"].value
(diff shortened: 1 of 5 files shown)


repos/exawind/packages/amr-wind/package.py
```diff
@@ -56,6 +59,10 @@ def cmake_args(self):
         if spec.satisfies("dev_path=*"):
             cmake_options.append(self.define("CMAKE_EXPORT_COMPILE_COMMANDS", True))
 
+        if "+umpire" in self.spec:
```
jrood-nrel, line 62: I think this `if` statement can be removed since umpire is only used in hypre.

repos/exawind/packages/exawind/package.py
```diff
@@ -88,6 +92,10 @@ def cmake_args(self):
 
         args.append(self.define("MPI_HOME", spec["mpi"].prefix))
 
+        if "+umpire" in self.spec:
```
jrood-nrel, line 95: I think this `if` statement can be removed since umpire is only used in hypre.

repos/exawind/packages/nalu-wind/package.py
```diff
@@ -95,6 +98,10 @@ def cmake_args(self):
             cmake_options.append(self.define("CMAKE_EXPORT_COMPILE_COMMANDS",True))
             cmake_options.append(self.define("ENABLE_TESTS", True))
 
+        if "+umpire" in self.spec:
```
jrood-nrel: I think this `if` statement can be removed since umpire is only used in hypre.

repos/exawind/packages/hypre/package.py
```diff
@@ -43,10 +43,5 @@ def configure_args(self):
         if "+umpire" in spec:
             if  (("+cuda" in spec or "+rocm" in spec) and "--enable-device-memory-pool" in options):
                 options.remove("--enable-device-memory-pool")
-            options.append("--with-umpire")
```
jrood-nrel, line 46: It seems like these should be necessary?

PaulMullowney: the main spack hypre package has the correct options in place.

jrood-nrel: Ah, great.

repos/exawind/packages/amr-wind/package.py
PaulMullowney: I've found that you need to compile with UMPIRE_DIR to properly resolve headers.

repos/exawind/packages/exawind/package.py
PaulMullowney: see above

repos/exawind/packages/nalu-wind/package.py
PaulMullowney: see above

repos/exawind/packages/amr-wind/package.py
PaulMullowney: To say it another way, if you don't do find package umpire in each of exawind, nalu-wind, and amr-wind, the compilation breaks because it can't find the umpire header files.

jrood-nrel: I guess it's because hypre uses autotools. If it used cmake, we wouldn't have to do this. Then we have to add the logic to Exawind and Nalu-Wind too.

jrood-nrel (approved):

psakievich (approved):

PR #494, opened 2023-02-08
Allow mixed device configuration with hypre

jrood-nrel:
This seems to work without needing a fork of both hypre and Nalu-Wind.

### DIFF
--- a/configs/base/packages.yaml
+++ b/configs/base/packages.yaml
@@ -25,7 +25,10 @@ packages:
   tioga:
     version: [develop]
   hypre:
-    version: [develop]
+    require: "@develop"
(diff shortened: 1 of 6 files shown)


psakievich (commented):
So what is it that triggers the symbol mangling for `hypre2`?

jrood-nrel:
Symbol mangling is unnecessary in this case. Thank goodness.

psakievich (approved):

jrood-nrel:
Ok I'm going to merge this. I haven't tested every configuration, but I will fix it if it somehow breaks concretization of anything else. So now anybody should be able to build for mixed device.

PR #496, opened 2023-02-10
Tiny Profile Variants

PaulMullowney:
Adding tiny profile variants


psakievich:
This was already added in mainline spack: https://github.com/spack/spack/pull/34993. So we shouldn't need the amr-wind one.

psakievich (changes requested):
Let's use the Tiny Profile variant added in https://github.com/spack/spack/pull/34993

PaulMullowney:
> Let's use the Tiny Profile variant added in [spack/spack#34993](https://github.com/spack/spack/pull/34993)

Ok, but should it propagate from exawind?

psakievich:
I'm fine with having it as an exawind variant. No objections from me there.

psakievich (approved):

PR #498, opened 2023-02-10
Get compile_commands.json working for tioga

psakievich:
Not sure why the old syntax didn't work, but it must not have worked since @PaulMullowney hit an error with the file not exisiting

### DIFF
--- a/repos/exawind/packages/tioga/package.py
+++ b/repos/exawind/packages/tioga/package.py
@@ -21,7 +21,7 @@ class Tioga(bTioga, SMCMakeExtension):
         options = super(Tioga, self).cmake_args()
 
         if spec.satisfies("dev_path=*"):
-            cmake_options.append(self.define("CMAKE_EXPORT_COMPILE_COMMANDS",True))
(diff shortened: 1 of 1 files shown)


repos/exawind/packages/tioga/package.py
```diff
@@ -20,8 +20,8 @@ def cmake_args(self):
         spec = self.spec
         options = super(Tioga, self).cmake_args()
 
-        if "dev_path" in spec:
-            options.append(self.define("CMAKE_EXPORT_COMPILE_COMMANDS",True))
+        if spec.satisfies("dev_path=*"):
+            cmake_options.append(self.define("CMAKE_EXPORT_COMPILE_COMMANDS",True))
```
PaulMullowney: Is this right? Should it be options?

psakievich: You're right. I just copied it from the nalu package and didn't check it.
```suggestion
            options.append(self.define("CMAKE_EXPORT_COMPILE_COMMANDS",True))
```

PaulMullowney (approved):

tasmith4 (approved):

PR #506, opened 2023-03-03
Pin Trilinos to a working version before Kokkos 4.0

jrood-nrel:
This seems to work for me on Crusher.


psakievich:
Should we just do this for crusher or should we do it everywhere? I'm okay with either I suppose

PaulMullowney (approved):

jrood-nrel:
I think we should just move to this commit entirely for now.

PR #526, opened 2023-05-04
Add ROSCO package

ndevelder:
ROSCO is a controls library commonly used with OpenFAST. This adds a package allowing user to build ROSCO. When loaded, package gives the env variable $ROSCO_DISCON which can be echoed and copied into ServoDyn.

### DIFF
--- a/repos/exawind/packages/openfast/package.py
+++ b/repos/exawind/packages/openfast/package.py
@@ -14,7 +14,7 @@ class Openfast(bOpenfast):
     patch("segfault_message.patch", when="%clang@12.0.1 build_type=RelWithDebInfo")
     version("fsi", git="https://github.com/gantech/openfast.git", branch="f/br_fsi_2")
 
-    variant("rosco", default=True,
+    variant("rosco", default=False,
             description="Build ROSCO controller alongside OpenFAST")
 
     depends_on("rosco", when="+rosco")
--- a/repos/exawind/packages/openfast/package.py
+++ b/repos/exawind/packages/openfast/package.py
@@ -14,6 +14,11 @@ class Openfast(bOpenfast):
     patch("segfault_message.patch", when="%clang@12.0.1 build_type=RelWithDebInfo")
     version("fsi", git="https://github.com/gantech/openfast.git", branch="f/br_fsi_2")
 
+    variant("rosco", default=True,
+            description="Build ROSCO controller alongside OpenFAST")
+
+    depends_on("rosco", when="+rosco")
+
     #def setup_build_environment(self, env):
     #    spec = self.spec
     #    machine = find_machine(verbose=False, full_machine_name=False)
--- a/repos/exawind/packages/rosco/package.py
+++ b/repos/exawind/packages/rosco/package.py
@@ -47,7 +47,6 @@ class Rosco(CMakePackage):
         options.extend(
             [
                 self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
-                self.define("CMAKE_Fortran_FLAGS", "-ffree-line-length-0"),
                 self.define_from_variant("CMAKE_POSITION_INDEPENDENT_CODE", "pic"),
             ]
         )
@@ -57,3 +56,12 @@ class Rosco(CMakePackage):
     def setup_run_environment(self, env):
         env.set('ROSCO_DISCON', self.prefix.lib + "/libdiscon.so")
         env.set('ROSCO_DISCON_DIR', self.prefix.lib)
+
+    def flag_handler(self, name, flags):
+        spec = self.spec
+        is_gcc = spec.compiler.name in ["gcc"]
+
+        if is_gcc:
+            flags.append("-ffree-line-length-0")
+        
+        return (flags, None, None)
--- a/repos/exawind/packages/rosco/package.py
+++ b/repos/exawind/packages/rosco/package.py
@@ -7,9 +7,11 @@ from spack.package import *
 
 
 class Rosco(CMakePackage):
-    """ROSCO controls package for OpenFAST from NREL"""
-    # Note: this only builds ROSCO controls library for inclusion with OpenFAST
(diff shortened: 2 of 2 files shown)


repos/exawind/packages/rosco/package.py
```diff
@@ -0,0 +1,57 @@
...
+#
+# SPDX-License-Identifier: (Apache-2.0 OR MIT)
+
+from spack.package import *
+
+
+class Rosco(CMakePackage):
+    """ROSCO controls package for OpenFAST from NREL"""
+    # Note: this only builds ROSCO controls library for inclusion with OpenFAST
+    # If the toolbox or tuning scripts are needed, please build manually
```
psakievich:
```suggestion
    """
    ROSCO controls package for OpenFAST from NREL
    Note: this only builds ROSCO controls library for inclusion with OpenFAST
    If the toolbox or tuning scripts are needed, please build manually
    """
```

psakievich: I would embed this note in the class description. It will then show up when people run `spack info rosco`. It seems like pertinent info for the end users who would be installing this.

repos/exawind/packages/rosco/package.py
```diff
@@ -0,0 +1,57 @@
...
+    version("2.4.0", tag="v2.4.0")
+    version("2.3.0", tag="v2.3.0")
+    version("2.2.0", tag="v2.2.0")
+
+
+    variant("shared", default=False, description="Build shared libraries")
+    variant("pic", default=False, description="Position independent code")
+
+    # Dependencies for OpenFAST Fortran
+    # depends_on("openfast")
```
psakievich, line 38: If this only for running with openfast then it probably makes sense to add it as a dependency. Perhaps just a [runtime dependency](https://spack.readthedocs.io/en/latest/packaging_guide.html#dependency-types) though?

psakievich (commented):
Looks good to me. I think this could be a target for builtin spack though. Seems pretty straightforward. @jrood-nrel what do you think?

repos/exawind/packages/rosco/package.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def cmake_args(self):
+        spec = self.spec
+
+        options = []
+
+        options.extend(
+            [
+                self.define_from_variant("BUILD_SHARED_LIBS", "shared"),
+                self.define("CMAKE_Fortran_FLAGS", "-ffree-line-length-0"),
```
jrood-nrel: I think this is a GCC specific flag so you probably want it only for `%gcc`. I also think it might be best to use Spack's flag handlers for this stuff like we do in Trilinos.

repos/exawind/packages/rosco/package.py
ndevelder: Yeah, I was conflicted on this one...it's really only supplying a library that openfast uses...so never runs per se...and definitely no build dependency. If we ever add in the python tools part of rosco then there'd be dependencies on openfast, python, and maybe a handful of python packages, but right now this is really only for getting the controls library auto-built along with openfast. Maybe the dependency should go the other way, from the openfast package with a "rosco" variant?

repos/exawind/packages/rosco/package.py
ndevelder: I wondered about this...I took that directly from ROSCO's setup.py script and didn't question. I added flag handling in openfast for openmp so I think it'd be similarly straightforward...but maybe @dzalkind could comment on this flag?

jrood-nrel: https://github.com/spack/spack/blob/16613408e4047542c61322ca7c165fea695f6d4c/var/spack/repos/builtin/packages/trilinos/package.py#L457-L482

repos/exawind/packages/rosco/package.py
psakievich: yeah that makes sense.  Want to add a variant to the openfast package in spack-manager as part of this PR?

ndevelder: ok, this should be in now...I'll test quick but it's pretty simple addition

repos/exawind/packages/openfast/package.py
```diff
@@ -14,6 +14,11 @@ class Openfast(bOpenfast):
     patch("segfault_message.patch", when="%clang@12.0.1 build_type=RelWithDebInfo")
     version("fsi", git="https://github.com/gantech/openfast.git", branch="f/br_fsi_2")
 
+    variant("rosco", default=True,
```
psakievich:
```suggestion
    variant("rosco", default=False,
```
Let's keep the default off.  Otherwise it may trigger rebuilds and other issues for people not using ROSCO

ndevelder: Fixed this! Makes sense.

repos/exawind/packages/rosco/package.py
ndevelder: I added flag handler...still a little unsure of whether it should be in "fflags" in the env flags or directly in the compiler inject flags? Thoughts?

ndevelder: And I should say it still builds fine with intel 2021, so I haven't broken anything by removing the flag

psakievich (approved):
LGTM

jrood-nrel (approved):

ndevelder:
Just as an FYI/small bit of documentation:

This can be added to your build in one of two ways:
1. Add "+rosco" to your openfast spec 
2. Add as a standalone spec i.e. "rosco@2.8.0"

ROSCO does not technically depend on or act as dependee in builds of any other package.

Standard behavior is that when this package is loaded in spack,or as a module, it provides an environment variable $ROSCO_DISCON that when echoed will print the location of libdiscon.so. If building with a custom lib name, there is another env variable $ROSCO_DISCON_DIR that will just print the lib directory.

PR #529, opened 2023-05-11
Update Spack and address several API changes

### LINKED ISSUE #528: Latest spack refactors Environment
Since [April 14](https://github.com/spack/spack/pull/36927) the following lines are causing issues when updating spack to the latest develop:

https://github.com/sandialabs/spack-manager/blob/485cc7ef214ce40521ad3ab697986f82d1f524bf/spack-scripting/scripting/cmd/manager_cmds/create_env.py#L49-L105

Perhaps some input from the spack team could help re-write them.

psakievich:
This is in response to https://github.com/spack/spack/pull/36927 ~~and the `try except` will likely only be temporary. We will likely need to scrub this once we get past the current round of application milestones.~~

Fixes for `spack manager create-env`:
- [x] unmanaged spack env's (-d) command
- [x] manager spack env's (-n) command
- [ ] tests that environments still get setup properly

Close #528 

Also have to address issues from https://github.com/spack/spack/pull/36273.


jrood-nrel (approved):

psakievich:
Ugh what a pain.  Need to re-think our concrete version strategy or start educating our developers to use the new syntax `[name]@=[version]`

spack-scripting/scripting/environment_utils.py
```diff
@@ -0,0 +1,35 @@
...
+# for more details.
+
+import spack.environment.environment as senv
+from spack.environment.environment import config_dict
+
+
+class SpackManagerEnvironmentManifest(senv.EnvironmentManifestFile):
+    """Spack-Manager extensiont to the manifest file for prototyping"""
+
+    def set_config_value(self, root, key, value):
```
psakievich, line 15: @alalazo this is a prototype of what we discussed in slack.  I'm not super happy with it though.  I feel like I'm going to need multiple functions, like the one below to add includes. It would be nice to craft a dictionary wrapper that will update both yaml contents and set the changed flag at the same time.  Curious to what your thoughts are 🤔

spack-scripting/scripting/cmd/manager_cmds/create_env.py
```diff
@@ -37,31 +37,15 @@ def create_env(parser, args):
...
+    environment.create_in_dir(theDir, init_file=args.yaml, keep_relative=True)
+
+    manifest = SpackManagerEnvironmentManifest(theDir)
 
-    if not args.yaml or not _unify_already_set(yaml):
-        yaml["spack"]["concretizer"] = {"unify": True}
-        with env.write_transaction():
-            env.write()
+    if not args.yaml:
+        manifest.set_config_value("concretizer", "unify", True)
```
psakievich, line 48: @alalazo here are the extension functions in action. I'm also thinking a syntax like what we use for `spack config add` might be nice. Something like...
```python
manifest.add_config("concretizer:unify:true")
# or
manifest.add_config("concretizer:unify", True)
```

PR #547, opened 2023-07-10
Update spack submodule for OpenFAST versions added

psakievich:


jrood-nrel:
I think we had to revert updating Spack last time, but let's try it.

jrood-nrel (approved):

PR #550, opened 2023-07-13
Netcdf-c requires for version

psakievich:


jrood-nrel:
Machine-specific configs like Summit will need `require` with this change.

jrood-nrel (approved):

PR #555, opened 2023-08-03
Update eagle

jrood-nrel:


psakievich:
Seems like this will break building on a compute node.  Is that correct?

jrood-nrel:
Yes but I think that's better than having it broke on the login nodes.

psakievich:
That's fine with me. Let's just broadcast that and give the users a chance to appeal first.

psakievich (approved):

PR #556, opened 2023-08-16
GPU-aware MPI additions for Frontier

jrood-nrel:
I'm not sure if we want these since it makes is slower, but since I put it all together it might be worth getting it into a mergable state.


repos/exawind/packages/amr-wind/package.py
```diff
@@ -41,6 +43,21 @@ def setup_build_environment(self, env):
             env.set("LSAN_OPTIONS", "suppressions={0}".format(join_path(self.package_dir, "sup.asan")))
         if "%intel" in self.spec:
             env.append_flags("CXXFLAGS", "-no-ipo")
+        if "+mpi" in self.spec and "+gpu-aware-mpi" in self.spec and "+rocm" in self.spec and find_machine(verbose=False, full_machine_name=False) == "frontier":
```
psakievich: Minor nitpick, but if we `require('+mpi', when='+gpu-aware-mpi')` then we don't need to check both.

psakievich (approved):

psakievich:
I think we should have this, but default it to off.

repos/exawind/packages/amr-wind/package.py
jrood-nrel: I agree.

jrood-nrel:
For posterity, I should also link to the documentation I was following for this: https://docs.olcf.ornl.gov/systems/frontier_user_guide.html#gpu-aware-mpi

repos/exawind/packages/amr-wind/package.py
psakievich: we should also add `conflicts("+gpu-aware-mpi", when="~cuda~rocm")`

psakievich: kind of pedantic but would be good for making sure the concretizer doesn't screw us over with something weird for debug cpu only builds on gpu machines

marchdf:
@jrood-nrel and @psakievich can we merge this? I am actually using it ;)

psakievich:
@marchdf I made a few changes based on my previous comments. Want to test them out for me really quick? 😉  Looks like something is up with CI too, but I don't think that is related to my changes. I'll investigate.

marchdf:
Yes I can test that now!

repos/exawind/packages/amr-wind/package.py
```diff
@@ -35,13 +37,32 @@ class AmrWind(SMCMakeExtension, bAmrWind):
     depends_on("hypre+umpire", when="+umpire")
     depends_on("hypre+sycl", when="+sycl")
     depends_on("hypre+gpu-aware-mpi", when="+gpu-aware-mpi")
+    depends_on("+mpi", when="+gpu-aware-mpi")
```
marchdf: `+mpi` should be `mpi`

marchdf:
with the change I suggest above, this builds on frontier.

PR #560, opened 2023-08-30
CEE: Force configs for cuda builds

psakievich:
Force the configs to give us the cuda builds we want. Does use an openmpi build with gcc@10.3.0 with a gcc@9.3.0 stack. Testing to see if that is an issue but this is the fastest way forward.


alanw0 (approved):
Looks good to me.
I'll try it by pulling and checking out this branch.

PR #562, opened 2023-08-31
Fix perlmutter/yaml

ajpowelsnl:
This PR corrects the xpmem spec (in `configs/perlmutter/compilers.yaml`) and removes Trilinos sha that could not be untar'ed, to use `develop` instead (in `configs/perlmutter/packages.yaml`).


psakievich:
@ajpowelsnl we are not currently using `trilinos@develop` there is some work that needs to be done to support it.  We are pinned to a specific commit of `trilinos` to get our challenge problem done.  I don't think we are in a position right now to support build errors off of the develop branch and won't be until we get our other milestones wrapped up.

scripts/create-exawind-snapshot.sh
```diff
@@ -51,7 +51,7 @@ elif [[ "${SPACK_MANAGER_MACHINE}" == "summit" ]]; then
   cmd "nice -n19 spack manager snapshot -m -s exawind%gcc+hypre+cuda+amr_wind_gpu+nalu_wind_gpu exawind%gcc+hypre~cuda"
 elif [[ "${SPACK_MANAGER_MACHINE}" == "perlmutter" ]]; then
   NUM_CORES=8
-  cmd "nice -n19 spack manager snapshot -m -s exawind%gcc+hypre+cuda+amr_wind_gpu+nalu_wind_gpu"
+  cmd "nice -n19 spack -d manager snapshot -m -s exawind%gcc+hypre+cuda+amr_wind_gpu+nalu_wind_gpu"
```
psakievich: probably don't want to leave the `-d` in there.
```suggestion
  cmd "nice -n19 spack manager snapshot -m -s exawind%gcc+hypre+cuda+amr_wind_gpu+nalu_wind_gpu"
```

psakievich (approved):

scripts/create-exawind-snapshot.sh
ajpowelsnl: Sounds good.

PR #565, opened 2023-09-12
Use Cray fortran to build OpenFAST

jrood-nrel:
Just in case it helps run the challenge problem.


psakievich (approved):
The number of modules requires scares me a bit here.

jrood-nrel:
This will most likely be abandoned. I don't think we need it.

PR #566, opened 2023-09-28
Fix for flag handler 

ndevelder:
I think this should fix the gcc fortran flag. I'll test and report back.


ndevelder:
This seems to work fine with gcc 10.3.0 so I think is good to go!

psakievich (approved):

PR #574, opened 2023-10-16
Prefer exawind@master

jrood-nrel:


psakievich (approved):

alanw0:
std::filesystem is new in C++17, could there be a missing compiler-flag somewhere that sets the c++ standard?

jrood-nrel:
We always set it. This used to work until Frontier came back from maintenance. It's like the AMD Clang doesn't have `<filesystem>` anymore. The `fgradient` application is the only thing that needs it so I'm just shutting off the amr-wind tests on Frontier. I dealt with a similar issue here https://github.com/AMReX-Codes/amrex/pull/3523 .

https://github.com/Exawind/amr-wind/blob/49cc3f20946f40851ebaff84c47773875e976543/CMakeLists.txt#L49

PR #575, opened 2023-11-03
Container for (spack-manager) CUDA GPU Build of Exawind  for NERSC Science Platform

ajpowelsnl:
The proposed code specifies a spack-manager-based GPU-capable container on NERSC science platforms.

- Use [NERSC Container Technology](https://docs.nersc.gov/development/podman-hpc/overview/)

## Build

```
podman-hpc build --no-cache -t <TAG_NAME> -f Dockerfile-containergpucuda .

```

## Run

```
podman-hpc run --rm --gpu -it <TAG_NAME>

```

## Expected Output

```
root@8953e3033a99:/exawind-entry/spack-manager# which exawind
/exawind-entry/spack-manager/snapshots/exawind/containergpucuda/2023-11-03/opt/linux-ubuntu22.04-zen3/gcc-11.4.0/exawind-git.d3c1aa4656fc3c6eccaec8c684671c82a3895172=multiphase-fyborjcfgnruxcks2vwzrc4guyw5toew/bin/exawind
root@8953e3033a99:/exawind-entry/spack-manager# exawind --help
usage: exawind [--awind NPROCS] [--nwind NPROCS] input_file
	-h,--help		Show this help message
	--awind NPROCS		Number of ranks for AMR-Wind (default = all ranks)
	--nwind NPROCS		Number of ranks for Nalu-Wind (default = all ranks)
```

## Helpful Hints
- After exiting a container, you must build the container again using a new tag name. 
- If using non-NERSC machines, Docker can be used instead of `podman-hpc`
- Be aware that `spack-manager` depends on a pinned version of Spack, a point of configure / build / runtime fragility

### DIFF
--- a/spack-scripting/scripting/cmd/manager_cmds/find_machine.py
+++ b/spack-scripting/scripting/cmd/manager_cmds/find_machine.py
@@ -97,6 +97,9 @@ machine_list = {
     "perlmutter": MachineData(
         lambda: os.environ["NERSC_HOST"] == "perlmutter", "perlmutter-p1.nersc.gov"
     ),
+    "containergpucuda": MachineData(
+        lambda: os.environ["CONTAINER_BUILD"] == "gpucuda", "containgpucuda.nodomain.gov"
+    ),
     # General
     "darwin": MachineData(lambda: sys.platform == "darwin", "darwin.nodomain.gov"),
 }
--- a/configs/containergpucuda/compilers.yaml
+++ b/configs/containergpucuda/compilers.yaml
@@ -1,6 +1,6 @@
 compilers:
 - compiler:
-    spec: nvcc@=12.2.91
+    spec: gcc@11.4.0 
     paths:
       cc: /usr/bin/gcc
       cxx: /usr/bin/g++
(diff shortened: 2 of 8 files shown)


spack
psakievich, line 1: This looks like it was an update to the submodule file, and not the spack commit?  Is that right?  We have a mirror only policy on spack changes so these changes would need to go into mainline spack.

spack-scripting/scripting/cmd/manager_cmds/find_machine.py
```diff
@@ -97,6 +97,9 @@ def is_e4s():
     "perlmutter": MachineData(
         lambda: os.environ["NERSC_HOST"] == "perlmutter", "perlmutter-p1.nersc.gov"
     ),
+    "containergpucuda": MachineData(
```
psakievich, line 100: I am not sure I like this name. Would we expect this to build any container using cuda, or specifically containers on perlmutter?  I would prefer to start with a more precise name and relax it rather than vice-versa.

psakievich (changes requested):
@ajpowelsnl thanks for doing this and sorry it has taken so long to get a review going.

PR #578, opened 2023-11-14
adding helics support

mbkuhn:


mbkuhn:
@jrood-nrel this implementation was sufficient for amr-wind+helics to compile locally with apple-clang. Do you need me to check on another machine/compiler? Another thing, some of the helics interface in AMR-Wind has been deprecated since it was put into the code, which means that the latest version, helics 3.4.0, does not compile, but the previous version, helics 3.3.2, does. I'm not sure how to put the preferred version into the spack info or into the spack-manager behavior, if possible.

jrood-nrel:
Use `depends_on("helics@:3.3.2", when="+helics")`. Then we can merge this.

mbkuhn:
this last commit fixes the helics versioning and also addresses Phil's comment on including +mpi automatically.

psakievich (approved):

PR #585, opened 2023-12-21
Add amrex roctx profiling option

marchdf:


psakievich (changes requested):
Let's add a `requires("+rocm", when="+roctex")`

psakievich (approved):

PR #591, opened 2024-02-05
set up waves2amr option for spec

mbkuhn:
* turn on enable_w2a
* define fftw directory
* declare fftw dependence


mbkuhn:
Waves2AMR is intended to be a submodule of AMR-Wind that is only compiled on demand (currently a PR for AMR-Wind). It requires FFTW, though, so this update to spack-manager does two things: turns on compile flags and declares the dependence on FFTW.

psakievich (approved):

PR #635, opened 2025-08-04
Add distribution cmd

jnhealy2:
Add `distribution` cmd to `spack-manager` to enable the bundling of a `spack environment` into a self contained source package distribution.

There are still some issues with this PR:

- Needs a way to specify aspects of the environment that should not be bundled
- Possibly needs a way to add additional attributes to the generated environment
- Copy of the spack installation needs to filter out sections of the spack tree containing user-specific content

### DIFF
--- a/docs/additional_commands/distribution/distribution.md
+++ b/docs/additional_commands/distribution/distribution.md
@@ -1,0 +1,35 @@
+# Distribution Command
+
+The distribution command bundles an active Spack environment into a self-contained higherarchy.  When run, the command selects the necessary components of a Spack enviornment to retain so that the resultant bundle can be re-installed on another machine with out the need to acquire any outside resources.  An exact copy of the exact version of Spack used to create the distribution bundle as well as a bootstrap mirror, any installed Spack extensions, package repositories, and a source mirror are included alongsied a configured spack environment. The resultant hierarchy looks like this:
+```
+.
+|- distro
+    |- bootstrap-mirror
+    |- environment
+    |- extensions
+    |- spack
+    |- spack_repo
+```
+
+## Modifying the generated Spack envrionment
+The `distribution` command includes optional behavior that allows the user to modify the generated Spack environment by including/excluding sections from the `spack.yaml` that is being sourced from the active environment. This is achieved by passing a directory containing relavent Spack yaml files to either include or exclude. For example `/path/to/excludes/pacakges.yaml` would be provided via `--exclude /path/to/excludes` and might look like this:
+
+```yaml
+packages: 
+    cxx: {}
+    all:
+        providers:
+            lapack: [netlib-lapack]
+```
+The above yaml would prevent cxx from appearing in the list of packages entirely and would prevent netlib-lapack from appearing in the list of valid lapack providers.
+
+Similarly, directories passed with `--include` would be copied into the distribution and added to its environment in the `include` section.
+
+## Using the resultant pacakge
+To install the package that was created by `spack manager distribution` an end user just needs to source the included `setup-env.*` script from the included copy of Spack, activate the included environment, configure a desired compiler, and install.  For example:
+```bash
+. distro/spack/share/spack/setup-env.sh
+spack env activate distro/environment
+spack compiler find
+spack install
+```
--- a/docs/additional_commands/distribution/distribution.md
+++ b/docs/additional_commands/distribution/distribution.md
@@ -1,6 +1,6 @@
 # Distribution Command
 
-The distribution command bundles an active Spack environment into a self-contained higherarchy.  When run, the command selects the necessary components of a Spack enviornment to retain so that the resultant bundle can be re-installed on another machine with out the need to acquire any outside resources.  An exact copy of the exact version of Spack used to create the distribution bundle as well as a bootstrap mirror, any installed Spack extensions, package repositories, and a source mirror are included alongsied a configured spack environment. The resultant hierarchy looks like this:
+The distribution command bundles an active Spack environment into a self-contained hierarchy.  When run, the command selects the necessary components of a Spack environment to retain so that the resultant bundle can be re-installed on another machine without the need to acquire any outside resources.  An exact copy of the exact version of Spack used to create the distribution bundle, as a bootstrap mirror, any installed Spack extensions, all package repositories, and a source mirror are included alongside a configured Spack environment. The resultant hierarchy looks like this:
 ```
 .
 |- distro
@@ -12,7 +12,7 @@ The distribution command bundles an active Spack environment into a self-contain
 ```
 
 ## Modifying the generated Spack envrionment
-The `distribution` command includes optional behavior that allows the user to modify the generated Spack environment by including/excluding sections from the `spack.yaml` that is being sourced from the active environment. This is achieved by passing a directory containing relavent Spack yaml files to either include or exclude. For example `/path/to/excludes/pacakges.yaml` would be provided via `--exclude /path/to/excludes` and might look like this:
+The `distribution` command includes optional behavior that allows the user to modify the generated Spack environment by including/excluding sections from the `spack.yaml` that is being sourced from the active environment. This is achieved by passing a directory containing relevant Spack YAML files to either include or exclude. For example `/path/to/excludes/packages.yaml` would be provided via `--exclude /path/to/excludes` and might look like this:
 
 ```yaml
 packages: 
@@ -21,12 +21,12 @@ packages:
         providers:
             lapack: [netlib-lapack]
 ```
-The above yaml would prevent cxx from appearing in the list of packages entirely and would prevent netlib-lapack from appearing in the list of valid lapack providers.
+The above YAML would prevent `cxx` from appearing in the list of packages entirely and would prevent `netlib-lapack` from appearing in the list of valid `lapack` providers.
 
 Similarly, directories passed with `--include` would be copied into the distribution and added to its environment in the `include` section.
 
 ## Using the resultant pacakge
-To install the package that was created by `spack manager distribution` an end user just needs to source the included `setup-env.*` script from the included copy of Spack, activate the included environment, configure a desired compiler, and install.  For example:
+To install the package that was created by `spack manager distribution` an end user needs to source the included `setup-env.*` script from the included copy of Spack, activate the included environment, configure a desired compiler, and install.  For example:
 ```bash
 . distro/spack/share/spack/setup-env.sh
 spack env activate distro/environment
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -37,7 +37,10 @@ def add_command(parser, command_dict):
     subparser.add_argument(
         "--filter-externals",
         action="store_true",
-        help="Remove any package settings that declare a path to an external installation of the package's binaries",
+        help=(
+            "Remove any package settings that declare a path to an external "
+            "installation of the package's binaries"
+        ),
     )
     subparser.add_argument(
         "--extra-data",
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -52,18 +52,18 @@ def add_command(parser, command_dict):
     command_dict["distribution"] = distribution
 
 
+def _read_config(filename):
+    return spack.config.read_config_file(filename)
+
+
 def get_env_as_dict(env):
-    return spack.config.read_config_file(os.path.join(env.path, "spack.yaml"))
+    return _read_config(os.path.join(env.path, "spack.yaml"))
 
 
-def get_local_config(name, path):
-    sections = list(spack.config.SECTION_SCHEMAS.keys())
-    scope = spack.config.DirectoryConfigScope(name, os.path.abspath(path))
-
+def get_local_config(path):
     little_config = {}
-    for section in sections:
-        if scope.get_section(section):
-            little_config.update(scope.get_section(section))
+    for f in os.listdir(path):
+        little_config.update(_read_config(os.path.join(path, f)))
     return little_config
 
 
@@ -102,19 +102,14 @@ def remove_subset_from_dict(larger_dict, subset_dict):
             if not value:
                 del larger_dict[key]
             elif isinstance(value, dict) and isinstance(larger_dict[key], dict):
-                # Recursively remove nested dictionaries
                 remove_subset_from_dict(larger_dict[key], value)
-                # If the nested dictionary becomes empty, remove the key
                 if not larger_dict[key]:
                     del larger_dict[key]
             elif isinstance(value, list) and isinstance(larger_dict[key], list):
-                # Remove matching items from the list
                 larger_dict[key] = [item for item in larger_dict[key] if item not in value]
-                # If the list becomes empty, remove the key
                 if not larger_dict[key]:
                     del larger_dict[key]
             elif larger_dict[key] == value:
-                # Remove the key if the value matches
                 del larger_dict[key]
     return larger_dict
 
@@ -132,7 +127,7 @@ def bundle_spack(location):
 
 class DistributionPackager:
     def __init__(self, env, root, includes=None, excludes=None, extra_data=None):
-        self.orig = env
+        self.environment_to_package = env
         self.includes = includes
         self.excludes = excludes
         self.extra_data = extra_data
@@ -147,6 +142,22 @@ class DistributionPackager:
         self._env = None
         self._cached_env = None
 
+    def __enter__(self):
+        self._cached_env = environment.active_environment()
+        tty.msg(f"Concretizing env: {self.environment_to_package.name}....")
+        self.environment_to_package.concretize()
+        self.environment_to_package.write()
+        environment.deactivate()
+        self.init_distro_dir()
+
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.filter_excludes()
+        self.remove_unwanted_artifacts()
+        if self._cached_env:
+            environment.activate(self._cached_env)
+
     @property
     def env(self):
         if self._env is None:
@@ -154,7 +165,7 @@ class DistributionPackager:
             self._env = environment.create_in_dir(epath, keep_relative=True)
         return self._env
 
-    def wipe_n_make(self):
+    def init_distro_dir(self):
         tty.msg("Precleaning....")
         if os.path.isdir(self.path):
             shutil.rmtree(self.path)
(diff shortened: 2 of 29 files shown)


jnhealy2:
@psakievich

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,240 @@
...
+    )
+    # TODO: Do not copy the site install config files
+    badfile = os.path.join(spack_install, "etc", "spack", "include.yaml")
+    if os.path.isfile(badfile):
+        os.remove(badfile)
+
+    # We deactivate the environment and use the special one we created so that the concretization will be the same.
+    #  Specifically, we do not want to omit any packages that are externals Just So They Build Faster internally, but are still needed externally.
+    #  However, this causes issues for packages that are not downloadable, so we do a first-shot mirror creation with the original environment active.
+    print(f"Creating mirror at {mirror}....")
```
psakievich: Print statements should be replaced with `tty.message` `tty.warning` or `tty.debug`. The `tty` module has parallel message support and a host of other features.

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,240 @@
...
+                    repos.add(
+                        os.path.normpath(
+                            os.path.join(
+                                os.path.dirname(
+                                    m.get_config_filename(scope=x, section="repos")
+                                ),
+                                repo,
+                            )
+                        )
+                    )
```
psakievich: I see this pattern a few times. Refactor into a method and add a unit-test for it?

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,240 @@
...
+
+def distribution(parser, args):
+    env = environment.active_environment()
+
+    if not env:
+        print(
+            "ERROR: You must have an active Spack environment to `spack distribution`.",
+            file=sys.stderr,
+        )
+        sys.exit(1)
```
psakievich:
```suggestion
    env = spack.cmd.require_active_env(cmd_name="distribution")
```

This does the same thing.

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,240 @@
...
+                    try:
+                        package_settings[package].update(data)
+                    except KeyError:
+                        package_settings[package] = data
+    # Trim out information about the MPI provider, since we can't know what the user will point at to build
+    # TODO: Decide how to handle the same issue for BLAS/LAPACK providers
+    remove_entries_with_substrings(package_settings, {"%", "mpi"})
+    print(f"Packing up Spack environment settings to {install_env}....")
+    os.makedirs(install_env)
+    big_config = {
```
psakievich: Spack-Manger has a functionality that already assembles a big config:
https://github.com/sandialabs/spack-manager/blob/main/manager/manager_cmds/include.py
and 
https://github.com/sandialabs/spack-manager/blob/main/manager/manager_cmds/includes_creator.py

It captures most of this minus the user specs. Might be worth refactoring both.

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,240 @@
...
+        if x.startswith("env") or x.startswith("include"):
+            packages = m.get("packages", scope=x)
+            for package, data in packages.items():
+                if "externals" not in data:
+                    try:
+                        package_settings[package].update(data)
+                    except KeyError:
+                        package_settings[package] = data
+    # Trim out information about the MPI provider, since we can't know what the user will point at to build
+    # TODO: Decide how to handle the same issue for BLAS/LAPACK providers
```
psakievich: Brainstorming here: Could we keep the specs, but delete the external information if they are there? If we are going for reproducibility we would want them to use the same mpi, blas and lapack.

``` yaml
packages:
  mpi: "openmpi"
  openmpi: "openmpi@4.0.4"
```

psakievich (changes requested):
Looks like a good start. The import changes we should revert. We should make `spack@1.0.1` the starting point for using this.

manager/manager_cmds/distribution.py
jnhealy2: Yes, that seems to be the sensible thing to do

manager/manager_cmds/distribution.py
jnhealy2: I don't know that we want to keep that much information.  There may be cases where the code being packaged wants to restrict the vendor/version of the MPI, but in many cases I think it makes more sense to let the user of the distribution define what MPI toolchain they want to use.  I think it might make sense to use a couple of input files that allow the user to specify specific sections of the environment they wish to remove as well as any specific configurations they explicitly want to include.

manager/manager_cmds/distribution.py
jnhealy2: Good to know, I'd been looking at Spack's API to see if there was a better way to handle this; sounds like we might already have the needed tools.

manager/manager_cmds/distribution.py
psakievich: Hmm yeah that is fair. As this evolves will you update the description with the API you are envisioning?

jnhealy2: I'm not really sure what a good approach might look like.  I'm currently thinking a `require/exclude.yaml` that defines any attributes the user would like to modify the generated env with...

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,378 @@
...
+        extra_data=args.extra_data,
+    )
+    with packager:
+        packager.configure_specs()
+        packager.configure_includes()
+        packager.configure_extensions()
+        packager.configure_package_repos()
+        packager.configure_package_settings()
+        packager.finalize()
+        packager.configure_package_mirror()
```
psakievich: rename to source mirror

jnhealy2: done

jnhealy2:
Currently `bundle_spack`, `configure_source_mirror`, and `configure_bootstrap_mirror` are all untested because of the cost of copying spack and the difficulty of mocking the mirror creation process.  This also means the main method is also untested.  Suggestions for addressing this are welcome.

manager/manager_cmds/distribution.py
jnhealy2: As of the latest change set have settled on user provided `include` and `exclude` directories that contain yaml files to either include or exclude.  For example `/path/to/excludes/pacakges.yaml` would be provided via `--exclude /path/to/excludes` and might look like this:
```yaml
packages: 
    cxx: {}
    all:
        providers:
            lapack: [netlib-lapack]
```
The above yaml would prevent `cxx` from appearing in the list of packages entirely and would prevent `netlib-lapack` from appearing in the list of valid `lapack` providers (the `all`, `providers`, and `lapack` keys would only be removed from the user's packages set if there were no other values in their respective dictionaries.

Similarly, directories passed with `--include` would be copied into the distribution and added to its environment as an `include`.

psakievich: Maybe we should have the command preserve externals and have this interface be how they get removed?

The thing I don't like about auto removing externals is that there is no clean intersection with `package.yaml` config to extract the intersection. 

So if users wanted to keep them, then they have to manually scrape them and add them back via an `include`.

I do see a case for auto removing all externals though. Perhaps that should just be a Boolean option (remove all or none).

psakievich:
for bundles spack IDK if we NEED a test. It's really just a method that calls the copy function you setup.

For the other two methods, the meat of the test is that the commands to the spack internal method are valid. Could we monkeypatch the calls and test them against spack's parser? I don't think we need to run the spack commands to test the function. That is tested in spack's release.

I think the real test is just that the API in spack doesn't break. In terms of what these functions do, it's not too much, and that's all we should test.

I think we can punt on these if it becomes too much work.

manager/manager_cmds/distribution.py
jnhealy2: The option is sounding like the best path at the moment.  I'll work-up what that might look like

jnhealy2:
Suggestions for testing were appreciated @psakievich, I've implemented the tests and I think the value proposition in there.

docs/additional_commands/distribution/distribution.md
```diff
@@ -0,0 +1,35 @@
+# Distribution Command
+
+The distribution command bundles an active Spack environment into a self-contained higherarchy.  When run, the command selects the necessary components of a Spack enviornment to retain so that the resultant bundle can be re-installed on another machine with out the need to acquire any outside resources.  An exact copy of the exact version of Spack used to create the distribution bundle as well as a bootstrap mirror, any installed Spack extensions, package repositories, and a source mirror are included alongsied a configured spack environment. The resultant hierarchy looks like this:
```
sebrowne:
```suggestion
The distribution command bundles an active Spack environment into a self-contained hierarchy.  When run, the command selects the necessary components of a Spack environment to retain so that the resultant bundle can be re-installed on another machine without the need to acquire any outside resources.  An exact copy of the exact version of Spack used to create the distribution bundle, as a bootstrap mirror, any installed Spack extensions, all package repositories, and a source mirror are included alongside a configured Spack environment. The resultant hierarchy looks like this:
```

docs/additional_commands/distribution/distribution.md
```diff
@@ -0,0 +1,35 @@
...
+|- distro
+    |- bootstrap-mirror
+    |- environment
+    |- extensions
+    |- spack
+    |- spack_repo
+```
+
+## Modifying the generated Spack envrionment
+The `distribution` command includes optional behavior that allows the user to modify the generated Spack environment by including/excluding sections from the `spack.yaml` that is being sourced from the active environment. This is achieved by passing a directory containing relavent Spack yaml files to either include or exclude. For example `/path/to/excludes/pacakges.yaml` would be provided via `--exclude /path/to/excludes` and might look like this:
```
sebrowne:
```suggestion
The `distribution` command includes optional behavior that allows the user to modify the generated Spack environment by including/excluding sections from the `spack.yaml` that is being sourced from the active environment. This is achieved by passing a directory containing relevant Spack YAML files to either include or exclude. For example `/path/to/excludes/packages.yaml` would be provided via `--exclude /path/to/excludes` and might look like this:
```

docs/additional_commands/distribution/distribution.md
```diff
@@ -0,0 +1,35 @@
...
+
+```yaml
+packages: 
+    cxx: {}
+    all:
+        providers:
+            lapack: [netlib-lapack]
+```
+The above yaml would prevent cxx from appearing in the list of packages entirely and would prevent netlib-lapack from appearing in the list of valid lapack providers.
```
sebrowne:
```suggestion
The above YAML would prevent `cxx` from appearing in the list of packages entirely and would prevent `netlib-lapack` from appearing in the list of valid `lapack` providers.
```

docs/additional_commands/distribution/distribution.md
```diff
@@ -0,0 +1,35 @@
...
+
+Similarly, directories passed with `--include` would be copied into the distribution and added to its environment in the `include` section.
+
+## Using the resultant pacakge
+To install the package that was created by `spack manager distribution` an end user just needs to source the included `setup-env.*` script from the included copy of Spack, activate the included environment, configure a desired compiler, and install.  For example:
```
sebrowne:
```suggestion
To install the package that was created by `spack manager distribution` an end user needs to source the included `setup-env.*` script from the included copy of Spack, activate the included environment, configure a desired compiler, and install.  For example:
```

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+
+def bundle_spack(location):
+    tty.msg(f"Packing up Spack installation to {location}....")
+    ignore_these = ["var/spack/environments/*", "opt/*", ".git*", "etc/spack/include.yaml"]
+    copy_files_excluding_pattern(spack_root, location, ignore_these)
+
+
+class DistributionPackager:
+    def __init__(self, env, root, includes=None, excludes=None, extra_data=None):
+        self.orig = env
```
sebrowne:
```suggestion
        self.environment_to_package = env
```

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+        self._cached_env = None
+
+    @property
+    def env(self):
+        if self._env is None:
+            epath = os.path.join(self.path, "environment")
+            self._env = environment.create_in_dir(epath, keep_relative=True)
+        return self._env
+
+    def wipe_n_make(self):
```
sebrowne:
```suggestion
    def init_distro_dir(self):
```

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+        tty.msg("Precleaning....")
+        if os.path.isdir(self.path):
+            shutil.rmtree(self.path)
+        os.makedirs(self.path)
+
+    def filter_excludes(self):
```
sebrowne, line 163:
```suggestion
    def construct_environment(self):
```

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+            tty.msg(f"Processing settings to exclude from env: {self.env.name}....")
+            cfg = get_local_config("exclude", self.excludes)
+            remove_subset_from_dict(env_data["spack"], cfg)
+        self._write(env_data)
+
+    def filter_excludes_and_concretize(self):
+        self.filter_excludes()
+        tty.msg(f"Concretizing env: {self.env.name}....")
+        self.env.concretize(force=True)
+        self.env.write()
```
sebrowne:
```suggestion
    def concretize(self):
        tty.msg(f"Concretizing env: {self.env.name}....")
        self.env.concretize(force=True)
        self.env.write()
```

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+        excludes=args.exclude,
+        extra_data=args.extra_data,
+    )
+    with packager:
+        packager.configure_specs()
+        packager.configure_includes()
+        packager.configure_extensions()
+        packager.configure_package_repos()
+        packager.configure_package_settings(filter_externals=args.filter_externals)
+        packager.filter_excludes_and_concretize()
```
sebrowne:
```suggestion
        packager.filter_excludes()
        packager.concretize()
```

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+            remove_subset_from_dict(env_data["spack"], cfg)
+        self._write(env_data)
+
+    def filter_excludes_and_concretize(self):
+        self.filter_excludes()
+        tty.msg(f"Concretizing env: {self.env.name}....")
+        self.env.concretize(force=True)
+        self.env.write()
+
+    def clean(self):
```
sebrowne: I'm a teensy bit conflicted about calling this `clean()`, but I don't know that I can come up with something better.  What I mean is that it's not _wiping_, but `clean` kind of implies that.  It's really _cleaning up_, or perhaps _reset_ ing?

tests/test_distribution.py
```diff
@@ -0,0 +1,529 @@
...
+
+    dest = os.path.join(tmpdir.strpath, "dest")
+    ignore = ["bar/other.txt", "bing/bang/*"]
+    distribution.copy_files_excluding_pattern(root, dest, ignore)
+
+    results = []
+    for _, _, files in os.walk(dest):
+        assert "other.txt" not in files
+        results.extend(files)
+    assert len(results) == 3
```
sebrowne: What do you think of calling `other.txt` `bad.txt` to be more explicit?

tests/test_distribution.py
```diff
@@ -0,0 +1,529 @@
...
+    assert data == orig
+
+
+def test_remove_subset_from_dict():
+    data = {"a": 1, "b": {"c": 2, "d": 3}, "e": [1, 2, 3], "f": [1, 2]}
+    orig = data.copy()
+    subset = {"a": None, "b": {"c": None}, "e": [1], "f": [1, 2]}
+    expected = {"b": {"d": 3}, "e": [2, 3]}
+    distribution.remove_subset_from_dict(data, subset)
+    assert data != orig
```
sebrowne: Extraneous assert, the `data == expected` is sufficient

sebrowne (commented):
Could you add meaningful docstrings to the tests?

manager/manager_cmds/distribution.py
jnhealy2: `remove_unwanted_artifacts()`?

tests/test_distribution.py
jnhealy2: Good suggestion. I will do that

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+        environment.deactivate()
+        self.wipe_n_make()
+
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.filter_excludes()
+        self.clean()
+        if self._cached_env:
+            environment.activate(self._cached_env)
```
davi0011: I would recommend these kinds of methods be placed at the top of a class in general to support the knowledge of how a class is used at first glance. Okay with that being follow on .

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+    def filter_excludes_and_concretize(self):
+        self.filter_excludes()
+        tty.msg(f"Concretizing env: {self.env.name}....")
+        self.env.concretize(force=True)
+        self.env.write()
+
+    def clean(self):
+        for item in os.listdir(self.env.path):
+            fullname = os.path.join(self.env.path, item)
+            if "spack.yaml" in item:
```
davi0011, line 191: The more and more I see this referenced as a variable the more I think the spack.yaml should be a class which you are using methods to edit/parse/pass.

manager/manager_cmds/distribution.py
```diff
@@ -0,0 +1,382 @@
...
+def remove_subset_from_dict(larger_dict, subset_dict):
+    """
+    Removes the subset dictionary from the larger dictionary, including items in lists,
+    and deletes keys if the subset value is an empty dictionary.
+
+    :param larger_dict: The larger dictionary from which the subset will be removed.
+    :param subset_dict: The subset dictionary to remove.
+    :return: The modified larger dictionary.
+    """
+    for key, value in subset_dict.items():
```
davi0011, line 100: Honestly, I am starting to think the comments in the middle of the code here add almost nothing and instead make it harder to parse.

jnhealy2: I can remove those

manager/manager_cmds/distribution.py
jnhealy2: Agree that a consistent interface into the YAML is needed.  Long-term this will be something worth addressing.

manager/manager_cmds/distribution.py
jnhealy2: Will do.

manager/manager_cmds/distribution.py
sebrowne: Do you think that's better, or is `clean()` fine and I'm nitpicking?

manager/manager_cmds/distribution.py
jnhealy2: done.

manager/manager_cmds/distribution.py
jnhealy2: done

tests/test_distribution.py
jnhealy2: done.

manager/manager_cmds/distribution.py
jnhealy2: I went ahead and changed it; if you have a preference I'm fine with anything

jnhealy2:
> docst

agreed, and added

sebrowne (approved):

psakievich (approved):

PR #637, opened 2025-09-03
Add buildcache creation to distribution cmd

jnhealy2:
Adds ability to bundle a buildcache with a spack distribution. The command will default to creating both a binary and source mirror. Currently signing of binaries when pushing them to the cache will be disabled.

### DIFF
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -8,6 +8,7 @@ import spack.extensions
 import spack.llnl.util.tty as tty
 import spack.util.path
 import spack.util.spack_yaml
+import spack.verify
 from spack import environment
 from spack.cmd.mirror import create_mirror_for_all_specs, filter_externals
 from spack.llnl.util.filesystem import working_dir
@@ -47,6 +48,23 @@ def add_command(parser, command_dict):
         help=(
             "Directory where any aditional data to be copied "
             "into the root of the packaged distribution"
+        ),
+    )
+    group = subparser.add_mutually_exclusive_group()
+    group.add_argument(
+        "--source-only",
+        action="store_true",
+        help=(
+            "Only create a source mirror for packages in the environment. "
+            "Default is to create a source and binary mirror"
+        ),
+    )
+    group.add_argument(
+        "--binary-only",
+        action="store_true",
+        help=(
+            "Only create a binary mirror for packages in the environment. "
+            "Default is to create a source and binary mirror"
         ),
     )
     command_dict["distribution"] = distribution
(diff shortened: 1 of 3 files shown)


jnhealy2:
@sebrowne and @psakievich, please review

sebrowne (approved):

manager/manager_cmds/distribution.py
```diff
@@ -354,15 +389,42 @@ def _write(self, data):
...
+            )
+            tty.warn(verification)
+            args.source_only = True
+            break
+    if args.source_only and args.binary_only:
+        tty.error(
+            "Binary distribution requested, but the environment does not "
+            "include the necessary installed binary packages"
+        )
+        raise SystemExit(1)
```
psakievich:
```suggestion
        tty.die(
            "Binary distribution requested, but the environment does not "
            "include the necessary installed binary packages"
        )
```
Same thing, less code.

psakievich (approved):
LGTM.  A couple of thoughts, but not sure we NEED to address them here:

1. Should we be deactivating/untrusting all the other spack mirrors as part of the environment setup?
2. Should we also run `spack buildcache keys --install --trust` with the environment prior to bundling? Since the mirror is unsigned there is no need to propagate gpg keys.
3. Same for the buildcache index.

jnhealy2:
> LGTM. A couple of thoughts, but not sure we NEED to address them here:
> 
>     1. Should we be deactivating/untrusting all the other spack mirrors as part of the environment setup?
> 
>     2. Should we also run `spack buildcache keys --install --trust` with the environment prior to bundling? Since the mirror is unsigned there is no need to propagate gpg keys.
> 
>     3. Same for the buildcache index.

For 1, I was initially going to do something like that, but as I worked it felt less necessary.  Do you @psakievich have any thoughts about the approach here vs requiring the user of the bundle to do something like `--cache-only`?

2.  In the initial implementation I was avoiding the complexity of how to best implement usage of gpg keys (do we require the user to have keys set up, do we generate them if they don't exist, etc?)  I think it is wise to add the singing but I would welcome thoughts about the approach to using keys 

3. I'm not clear on what should/could be done for the buildcache index on this end

psakievich:
@jnhealy2 would you run the user side of this and document the steps for what had to be done to get a cache only build to work? An easy environment would be serial kokkos as a root spec since it has a small set of dependencies. 

My main line of thinking is to identify and minimize any generic commands outside of `spack install --cache-only` that are required upon activation. If none of these other commands are needed for the sierra use case then I'm okay with punting on gpg keys for this iteration.

jnhealy2:
I updated the process to trust the buildcache.  The user side process is as follows
```bash
source distro/spack/share/spack/setup-env.sh
spack env activate distro/environment
spack install --cache-only
```

PR #638, opened 2025-09-09
Fix repository order bug in distribution command

jnhealy2:
The distribution command was modifying the list of repos in a given spack environment which causes problems when resolving package dependencies. Maintain the original package repository data structure throughout the distro bundling process so that this does not occur.


jnhealy2:
@psakievich, this is the fix for the repo ordering issue

psakievich (approved):

PR #639, opened 2025-10-16
Minor fixes

jnhealy2:
fix tests to be robust to non-breaking api changes. Also, change binary asset check so it is not recursive. This avoids an infinite loop in some environments.


jnhealy2:
@psakievich, this addresses the break with `spack develop` and fixes an issue I hit with Sierra when creating a binary distribution

psakievich (approved):

PR #640, opened 2025-10-29
Distribution extra spack cleanup

jnhealy2:
The distribution command was including files specific to the user's spack environment
were getting bundled along with spack.  This is because spack writes into its root
directory as it is being used.  This creates a more exhaustive list of locations
to avoid copying along with spack and more aggressively asserts they don't end up in
the final spack installation

### DIFF
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -1,4 +1,5 @@
 import fnmatch
+import glob
 import os
 import shutil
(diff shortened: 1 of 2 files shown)


jnhealy2:
@psakievich, this should do a better job being precise about our spack copy.  It looks to me like the CI had a systemic failure.

psakievich (approved):

PR #641, opened 2025-11-04
Add workers argument to create_mirror_for_all_specs function in distribution script

akimler:
A change in the develop branch of Spack has caused an error with our distribution.py script. The function `create_mirror_for_all_specs` now has a required argument "workers" that was previously not set. 

This PR adds the workers argument to be set to the job number defined by Spack by using the `determine_number_of_jobs` function in Spack config.


psakievich:
Since we have an objective to refactor this function to call the full `spack mirror` command with `spack.main.SpackCommand` I'm going to close this PR.

jnhealy2 (approved):

PR #645, opened 2026-01-27
Extract and use full flattened configuration for the initial creation of the distribution environment

akimler:
With this change the distribution process will use the fully flattened configuration from the targeted Spack environment. This change was necessary to make sure that we included all of the configuration file that is included in the Spack environment.

Some sections in the configuration have been excluded by default (include, mirrors, repos) and an exclude argument (option to use a file or a command line argument) was added for further customization.

### DIFF
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -44,7 +44,6 @@ def add_command(parser, command_dict):
         "--exclude-file",
         help="Sections in the enviroment's configuration file to exclude located a file",
     )
-    create
     subparser.add_argument(
         "--exclude-configs",
         nargs="+",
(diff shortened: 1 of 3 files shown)


manager/manager_cmds/distribution.py
```diff
@@ -219,6 +210,33 @@ def remove_unwanted_artifacts(self):
             else:
                 os.remove(fullname)
 
+    def get_flattened_config(self):
```
jnhealy2: It seems like the three methods you created are meant to be called in sequence and that they all need to be called (are one logical unit).  Would it make sense to make these three methods private and make a public interface that orchestrates calling them?

akimler: I just pushed a change that combined the three functions (get_flattened_config, create_config, and get_relative_path for extensions) into one function.

jnhealy2 (approved):

PR #646, opened 2026-01-29
Modify distribution sub-cmd to better support mirror creation

jnhealy2:
Mirror creation had several problems:

- The nature of what the distribution package does (Switching between two environments and constructing comparable states) means the bootstrap mirror needs to be registered in both environment before other mirrors are created.
- We need to be able to filter out specs that we don't want in the source mirror

Also, refactor usages of SpackCommand to call the desired method directly.


jnhealy2:
@sebrowne, can you please review?

sebrowne (approved):

PR #648, opened 2026-03-31
Remove buildcache keys call from distribution command

rtrigg:
In the distribution command, `configure_binary_mirror()` pushes with `--unsigned` but then calls `buildcache keys --install --trust`, which fails if no mirror is configured in the active environment's scope.

It exits before `bundle_spack()` runs, leaving an incomplete distro without the bundled Spack.

The keys call is unnecessary since there are no signatures to verify when pushing unsigned.

Tested with Spack 1.1.1.

```
  File "spack-manager/manager/manager_cmds/distribution.py", line 388, in configure_binary_mirror
    call(spack.cmd.buildcache, "buildcache", ["keys", "--install", "--trust"])
...
    spack.binary_distribution.get_keys(args.install, args.trust, args.force)
  File "spack/lib/spack/spack/binary_distribution.py", line 2213, in get_keys
    tty.die("Please add a spack mirror to allow " + "download of build caches.")
  File "spack/spack/lib/spack/spack/llnl/util/tty/__init__.py", line 249, in die
    sys.exit(1)
SystemExit: 1
```

### DIFF
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -385,7 +385,6 @@ class DistributionPackager:
         with self.environment_to_package:
             tty.msg(f"Creating binary mirror at {self.binary_mirror}....")
             call(spack.cmd.buildcache, "buildcache", ["push", "--unsigned", self.binary_mirror])
-            call(spack.cmd.buildcache, "buildcache", ["keys", "--install", "--trust"])
 
         mirror_path = os.path.join(
             os.path.relpath(self.path, self.env.path), os.path.basename(self.binary_mirror)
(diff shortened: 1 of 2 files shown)


rtrigg:
@jnhealy2 does this make sense? or should we instead 1) remove the --unsigned call from configure_binary_mirror, or should we 2) expose a flag that lets one use keys or not? Seems like a design decision.

jnhealy2:
> @jnhealy2 does this make sense? or should we instead 1) remove the --unsigned call from configure_binary_mirror, or should we 2) expose a flag that lets one use keys or not? Seems like a design decision.

Yeah, I think we had started implementing signed buildcache generation, and then put that down.  This makes sense to remove for now.

jnhealy2:
@rtrigg, looks like there is a unit test failing.  After its fixed we can get this merged.  Let me know if you want/need any help fixing it.

rtrigg:
Should be good now, thanks

jnhealy2 (approved):

PR #649, opened 2026-04-14
Distribution: Fix unable to bootstrap when an environment requires non-gcc/clang compilers

jnhealy2:
We ran into a problem creating a bootstrap mirror from an environment containing the following:
```
packages:
    c:
        require: intel-oneapi-compilers
    cxx:
        require: intel-oneapi-compilers
```
because clingo requires a gcc/clang compiler to build.  This change adds a fallback so that if the bootstrap mirror creation fails, it is re-attemepted in a temporary environemnt that does not contain any user-specified information.  This fix relies on spack finding a gcc compiler on the system that can be used to build clingo.

### DIFF
--- a/tests/test_distribution.py
+++ b/tests/test_distribution.py
@@ -915,8 +915,8 @@ def test_DistributionPackager_configure_bootstrap_mirror_fallback_to_empty_env(
 ):
     """
     Test that a failure creating the bootstrap mirror falls back to using an empty environment
-    (generally this has been caused by overly-restrictive compiler settings, which don't really apply
-    for the bootstrap mirror).
+    (generally this has been caused by overly-restrictive compiler settings, which don't really
+    apply for the bootstrap mirror).
     """
     pkgr, root, env = init_test_environment(tmpdir.strpath)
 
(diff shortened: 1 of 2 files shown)


tests/test_distribution.py
```diff
@@ -909,6 +910,37 @@ def __init__(self, _, cmd, args):
             parser.parse_args(call)
 
 
+def test_DistributionPackager_configure_bootstrap_mirror_fallback_to_empty_env(
+    tmpdir, monkeypatch
+):
+    """
+    This test verifies that `configure_bootstrap_mirror` does not construct a call to
+    `spack bootstrap` that violates its API.
```
sebrowne: Can you be a bit more specific?  I think this is valid, but it seems like the primary purpose of the test is more like `"""Test that a failure creating the bootstrap mirror falls back to using an empty environment (generally this has been caused by overly-restrictive compiler settings, which don't really apply for the bootstrap mirror)"""`

jnhealy2: done

sebrowne (approved):

PR #650, opened 2026-04-15
Preserve lockfile in distribution

rtrigg:
This adds the lockfile to the distribution command to make concretizing easy when the environment is installed from the distribution. Also updates a test that's only expecting one file (spack.yaml) to expect two with the spack.lock now.

### DIFF
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -240,7 +240,7 @@ class DistributionPackager:
                 remove_by_pattern(os.path.join(self.spack_dir, epattern), include_patterns)
         for item in os.listdir(self.env.path):
             fullname = os.path.join(self.env.path, item)
-            if "spack.yaml" in item:
+            if "spack.yaml" in item or "spack.lock" in item:
                 continue
             elif os.path.isdir(fullname):
                 shutil.rmtree(fullname)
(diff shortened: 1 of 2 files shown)


rtrigg:
Maybe I should add `assert "spack.lock" in env_assets` to the test as well? Also if we want to make this a `--lockfile` flag rather than the default, let me know and I'll pipe that through.

rtrigg:
I gotta fix some tests that mock only the bundle env and not the source env

rtrigg:
I pinned to spack v1.1.1 and I'm seeing passing tests (except for style, which is maybe because ruff isn't pinned?) https://github.com/rtrigg/spack-manager/actions/runs/25768143731/job/75685243341

rtrigg:
Just getting back to this, all the CI tests are now passing since the recent updates in main from @jnhealy2 

Open to any comments; if there's another way this should be done/tested, I'm open to that. The idea behind it was that if you are packaging up all the binaries for an environment, you shouldn't have to go through the work later to reconcretize. I've been using this on an existing project I've migrated to using `spack manager distribution` and that is working, fwiw.

jnhealy2:
This looks like a great optimization.  Sorry, I'm just getting to this @rtrigg.

jnhealy2 (approved):

PR #652, opened 2026-05-28
Update require_active_env to use subparser for latest Spack

jrood-nrel:


jrood-nrel:
Oh looks like maybe this is in https://github.com/sandialabs/spack-manager/pull/651

PR #653, opened 2026-06-23
Ensure relative path of boostrap config is able to be used for binary distribution creation

akimler:
During our use of the distribution.py extension we ran into a problem running on HPC compute nodes without access to the internet to grab the packages needed for bootstrapping for binary packages. This is resolved by using the source package's bootstrap config, but it does not account for relative paths for the config. This change allows for relative paths to exist with the bootstrap config used for the binary distribution creation.

### DIFF
--- a/manager/manager_cmds/develop.py
+++ b/manager/manager_cmds/develop.py
@@ -9,7 +9,10 @@ import os
 import shutil
 
 import spack.cmd
-import llnl.util.tty as tty
+try:
+    import llnl.util.tty as tty
+except ImportError:
+    import spack.llnl.util.tty as tty
 import spack.util.executable
 from spack.cmd.develop import develop as s_develop
 from spack.cmd.develop import setup_parser as s_setup_parser
--- a/manager/manager_cmds/develop.py
+++ b/manager/manager_cmds/develop.py
@@ -9,7 +9,10 @@ import os
 import shutil
 
 import spack.cmd
-import spack.llnl.util.tty as tty
+try:
+    import llnl.util.tty as tty
+except ImportError:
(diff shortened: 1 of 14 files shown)


tests/test_distribution.py
```diff
@@ -884,30 +893,33 @@ def __init__(self, _, cmd, args):
...
     # Create a fake bootstrap mirror structure on disk
-    # Structure: /tmp/fake_mirror/metadata/sources/...
     fake_mirror_root = os.path.join(tmpdir.strpath, "fake_mirror")
     metadata_dir = os.path.join(fake_mirror_root, "metadata")
-    sources_dir = os.path.join(metadata_dir, mirror_type)
-    os.makedirs(sources_dir)
-    with open(os.path.join(sources_dir, "some_data.yaml"), "w") as f:
+    metadata_subdir = os.path.join(metadata_dir, mirror_type)
+    os.makedirs(metadata_subdir, exist_ok=True)
+    with open(os.path.join(metadata_subdir, "some_data.yaml"), "w") as f:
```
sebrowne: Why change the name?  The original was more specific.

akimler: Yeah true, I changed this back to sources_dir

manager/manager_cmds/develop.py
```diff
@@ -9,7 +9,10 @@
 import shutil
 
 import spack.cmd
-import spack.llnl.util.tty as tty
+try:
+    import llnl.util.tty as tty
+except ImportError:
+    import spack.llnl.util.tty as tty
```
sebrowne, line 15: Why are these import changes (in all files) necessary?

PR #657, opened 2026-07-01
Revert "Preserve lockfile in distribution"

jnhealy2:
Reverts sandialabs/spack-manager#650

I had to revert this @rtrigg.  In my local testing it looked great, but I over looked a use-case.  When the distribution command is called with the `--filter-externals` command, the tool modifies the new environment.  When that happens we have to reconcretize in order to have those environment changes reflected in the final package that gets created.

### DIFF
--- a/manager/manager_cmds/distribution.py
+++ b/manager/manager_cmds/distribution.py
@@ -244,15 +244,9 @@ class DistributionPackager:
         os.makedirs(self.path)
 
     def concretize(self):
-        src_lock = self.environment_to_package.lock_path
-        dst_lock = self.env.lock_path
-        if os.path.isfile(src_lock):
-            tty.msg(f"Copying lockfile from {self.environment_to_package.name} to bundled env....")
-            shutil.copy2(src_lock, dst_lock)
-        else:
-            tty.msg(f"Concretizing env: {self.env.name}....")
-            self.env.concretize(force=True)
-            self.env.write()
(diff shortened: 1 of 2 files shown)


jnhealy2:
It would be possible to implement this behavior as optional, such that it does not impact the default workflow.

rtrigg:
Okay, I think there's a few ways to do this.

1) We just don't include the lockfile when --filter-externals is used. This is my vote; I think preserving the lockfile when possible should be a default behavior. To me that seems in line with what a user might reasonably expect.

2) We have an explicit flag like --keep-lockfile that conflicts with --filter-externals. I'm willing to do that, but whether it's worth the flag I can't answer. Alternatively, a --no-lockfile flag that is automatically set when using --filter-externals, but can be used on it's own. This keeps the default expectations from above, and might actually  be my favorite now that I think about it.

Let me know what you think and I'll open a new PR.

jnhealy2:
@rtrigg, I think option 1 sounds like a good starting point.  Technically we are significantly modifying the environment in the distribution code, so a re-concretization is always necessary, but the change you are proposing might work under some conditions.  This change has exposed an integration testing gap that exercises the end-to-end behavior of the distribution command.  If you re-propose the change and base it on --filter-externals, I will do some thorough manual testing of it.  We should be able to arrive at a workable solution.  When I have time I will try to write some integration tests to close the gap.

